PR #9, opened 2021-04-25
Support for http driver

d1bevz:
A few changes
Tested on clickhouse server ver. 19.16.10.44 with ReplicatedMergeTree engine table


d1bevz:
Well I tried... 
Seed not working due to zookeeper parameter

PR #11, opened 2021-07-19
Adds verify + secure boolean to ClickhouseCredentials

aaliang:
Allows users to specify clickhouse-driver Client options:
- `secure` if their ClickHouse instance is behind a secure port
- `verify` if ssl verify should be performed.

_(both defaulted to False)_


silentsokolov:
Thanks!

PR #15, opened 2021-08-27
First version of distributed materialization

Volodin-DD:
NOTE: lightly tested (though successfully). Need help from community.

Made two materializations: distributed_table for distributed version of table models and distributed_incremental for incremental models.

The main idea is to make local tables on cluster with engine and parameters from config in model's sql file within a distributed table "looking" on these local tables. So when referencing model with distributed materialization dbt will make insert into distributed table and select from distributed table. Sharding key is also to be set on model's config.

Main flow described in comments distributed_incremental.sql


Volodin-DD:
There were some issues with schemas. Fixed it

genzgd:
Closed in favor of #172 and #186

PR #17, opened 2021-09-28
Fix update and delete in snapshot.sql

oev81:


dbt/include/clickhouse/macros/materializations/snapshot.sql
```diff
@@ -108,60 +108,95 @@
...
+    select
+      'insert' as dbt_change_type,
+      source_data.*
+    from insertions_source_data as source_data
+    left outer join snapshotted_data on snapshotted_data.dbt_unique_key = source_data.dbt_unique_key
     where snapshotted_data.dbt_unique_key is null
-      or (snapshotted_data.dbt_unique_key is not null
-      and snapshotted_data.dbt_valid_to is null
+      or (
+        snapshotted_data.dbt_unique_key is not null
```
oev81, line 159:
```suggestion
      snapshotted_data.dbt_unique_key is not null
```

silentsokolov:
Thanks! Don't worry about error in tests. It's a inner dbt's problem and python.

guykoh:
Thanks for those changes - I implemented them in a separate PR where I also replaced the temporary table with MergeTree table.

Closing this PR now.

PR #18, opened 2021-10-01
Ability to run dbt tests with ClickHouse

artamoshin:
Solve of this issue was suggested in dbt-labs/dbt#3872.
Same method used in MS SQL adapter.


silentsokolov:
Thanks!

PR #37, opened 2022-04-24
Add support for Join engine

vlourme:
**Summary:**
This PR adds the support for `Join` engine, useful for joining large tables that couldn't be done using a `MergeTree()`.

---
**Discussion:**
In my opinion, we should add a way to disable the default `order by (tuple())` to avoid adding manually a support for each engine that does not support this instruction.

I have two proposals for this:
- ~~Removing the default value (`tuple()`), but this could break some existing code (?)~~ Breaking, the empty tuple is required for `MergeTree` ([see here](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree/#mergetree-query-clauses))
- Adding a support for `None` value, something such as:
  ```
  {{ config(
    engine='Join(ANY, LEFT, id)',
    order_by=None
  ) }}
  ```

If anyone has any better idea for implementing this, I'm open to discussion!

### DIFF
--- a/dbt/include/clickhouse/macros/adapters.sql
+++ b/dbt/include/clickhouse/macros/adapters.sql
@@ -35,7 +35,7 @@
       {%- if not loop.last -%},{%- endif -%}
     {%- endfor -%}
     )
-  {%- elif engine and engine.startswith('Distributed') -%}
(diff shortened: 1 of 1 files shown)


silentsokolov:
Thanks you!

PR #39, opened 2022-05-10
Add dbt pytest

guykoh:
Replacing the old deprecated dbt testing with new tests based on pytest following dbt documentation: https://docs.getdbt.com/docs/contributing/testing-a-new-adapter

Note: This pr is just setting up the testing configuration and writing down a simple base materialization test. Following this pr I'll add more tests and finally remove the old test folder.


silentsokolov:
Thanks for your PR. I appreciate it.

But I think:
- We have to stick to the original implementation. Docker must not be as part of the tests. You can note it in the README, but Docker can't be a dependency.
- The project must not to have two different testing ways. So I will merge it after PR is completed.

guykoh:
Thanks for the feedback.
See my responses below:
* The clickhouse server must be up for the tests to pass. We can mock the server but then we're not really run integration tests. Do you have other ideas for how to test it without docker? Do you expect the user to run clickhouse server every time they run the tests?
* I'll add the rest of the tests and remove the old ones from this pr.

guykoh:
Hi Dmitriy,
I updated my PR:
1. Added more tests and remove the old ones
2. Made docker-compose optional (by default it is set to false), and made sure there are no dependencies in docker.

Can you please review again and approve if it looks good to you?

dbt/adapters/clickhouse/impl.py
```diff
@@ -233,6 +233,23 @@ def get_csv_data(self, table):
...
+        try:
+            result = cursor.execute(sql)
+            if fetch == "one" and len(result) > 0:
+                return result[0]
+            elif fetch == "all":
+                return result
+            else:
+                return
+        except BaseException as e:
+            print(sql)
```
silentsokolov, line 247: Maybe we could use std logger. What do you think?

tests/integration/adapter/test_basic.py
```diff
@@ -0,0 +1,25 @@
...
+
+class TestEmptyClickHouseAdapter(BaseEmpty):
+    pass
+
+
+class TestIncrementalClickHouseAdapter(BaseIncremental):
+    pass
+
+
+class TestSnapshotTimestampMyAdapter(BaseSnapshotTimestamp):
```
silentsokolov, line 20: Maybe remove the prefixes `ClickHouseAdapter` / `MyAdapter` and just use simple names `BaseSnapshotTimestamp` -> `TestSnapshotTimestamp`, `BaseEmpty` -> `TestEmpty`.

silentsokolov (approved):

silentsokolov:
Thanks you!
I added couple comments and approved PR. You could merge if you not agree with these comments.

dbt/adapters/clickhouse/impl.py
guykoh: Good idea, I used AdapterLogger from dbt

tests/integration/adapter/test_basic.py
guykoh: Good idea - done

PR #41, opened 2022-05-16
Support 'engine=MergeTree()' in config

guykoh:


silentsokolov:
If we want to solve the problem forever. We must refuse `endswith` because there are CollapsingMergeTree, ReplacingMergeTree and etc with parameters `ver`/`sign`/etc. What do you think?

silentsokolov (approved):

PR #48, opened 2022-05-31
Cc with insert

guykoh:


dev_requirements.txt
```diff
@@ -1,5 +1,5 @@
 dbt-core==1.1.0
-clickhouse-connect==0.0.9
+clickhouse-connect>=0.0.12
```
genzgd, line 2: Let's update to 0.0.13, should be published shortly

setup.py
```diff
@@ -54,17 +54,16 @@ def _dbt_clickhouse_version():
     },
     install_requires=[
         f'dbt-core=={dbt_version}',
-        'clickhouse-driver>=0.2.2',
+        'clickhouse-connect>=0.0.12',
```
genzgd, line 57: Same -- 0.0.13

genzgd (changes requested):

setup.py
guykoh: done

genzgd (approved):

PR #56, opened 2022-06-09
Boolen field csv

guykoh:
Fixed support for boolean field csv seed.

The problem was the line terminator of the csv writer. By default it is set to /r/n and for some reason this confuses CH


genzgd (approved):
We should probably test against 21.8 ClickHouse since special CSV Bool behavior isn't documented and I don't know when it started working

genzgd:
Oh, I should have looked at the tests first.

PR #58, opened 2022-06-15
Fix: Table defaul.xxx doesn't exists.

1c7:


guykoh (approved):

guykoh:
Hi @1c7, thanks for this pr, I totally agree with this change. There's small workaround needed in order to make the tests pass. I added your change and fix the tests in this [pr](https://github.com/ClickHouse/dbt-clickhouse/pull/61) so I'm closing this pr now and will submit the other.

PR #77, opened 2022-07-11
Fix brackets regex in column.py

guykoh:


dbt/adapters/clickhouse/column.py
```diff
@@ -17,28 +17,26 @@ class ClickhouseColumn(Column):
         'INTEGER': 'Int64',
     }
     is_nullable: bool = False
-    _brackets_regex = re.compile(r'(Nullable|LowCardinality)\((.*?\))\)')
+    _brackets_regex = re.compile(r'^(Nullable|LowCardinality)\((.*?)\)$')
```
genzgd, line 20: Do we handle LowCardinality(Nullable(String)) types with this?

genzgd (approved):
LGTM, not sure the edge case of LowCardinality(Nullable) is actually important

yoavo-datricks:
Hi @guykoh. I have a feeling this wouldn't work with map columns.

guykoh:
> 

@yoavo-datricks, do you mean - https://clickhouse.com/docs/en/sql-reference/data-types/map? Why do you think it won't work?

yoavo-datricks:
Hi.
Yeah, I thought that the fact it has two types will not work for the regex but it looks good.
Thanks for merging this!

Issue #86, opened 2022-07-28
Using custom_settings with native driver gives an error


mgsrivastava:
I’m trying to add some dbt specific settings to `profiles.yml` using the custom_settings field. Running models works fine (and the settings are applied) when I use the http driver. But when using the native driver, I get the following error message:

```
Compilation Error in macro statement (macros/etc/statement.sql)
  __init__() got an unexpected keyword argument 'distributed_ddl_task_timeout'

  > in macro clickhouse__list_schemas (macros/adapters.sql)
  > called by macro list_schemas (macros/adapters/metadata.sql)
  > called by macro statement (macros/etc/statement.sql)
exit status 2
```

Running `dbt debug` shows similar behavior. All checks pass using the http driver, while the connection test fails using the native driver:

```
1 check failed:
dbt was unable to connect to the specified database.
The database returned the following error:

  >__init__() got an unexpected keyword argument 'distributed_ddl_task_timeout'
```

genzgd:
Yeah it's a bug.  Should be fixed in the next release with https://github.com/ClickHouse/dbt-clickhouse/pull/87

mgsrivastava:
Awesome, thanks!

guykoh:
https://github.com/ClickHouse/dbt-clickhouse/pull/87 suppose to fix that issue - closing this now

PR #87, opened 2022-07-28
Fix ClickHouse settings issue with native driver

genzgd:


genzgd:
Fixes https://github.com/ClickHouse/dbt-clickhouse/issues/86

guykoh (approved):

PR #92, opened 2022-09-01
Adaptive materialization strategy when using Atomic schema engine

Barsoomx:
When working with complex DBT projects, sometimes Cron jobs and builds interfere with each other, and cause SELECT errors or deadlocks when tables are being recreated by one or several processes. This PR, in combination with Atomic engine and [SYNC clickhouse setting](https://clickhouse.com/docs/en/operations/settings/settings/#database_atomic_wait_for_drop_and_detach_synchronously) should help avoid those issues.

For now table & view materializations are copied from dbt-core@v1.1.2 and modified

https://github.com/dbt-labs/dbt-core/tree/v1.1.2

When the 1.2.x support update is released I can update them accordingly

### DIFF
--- a/dbt/include/clickhouse/macros/materializations/table.sql
+++ b/dbt/include/clickhouse/macros/materializations/table.sql
@@ -1,4 +1,4 @@
-{% materialization table, default %}
+{% materialization table, adapter='clickhouse' %}
   {%- set identifier = model['alias'] -%}
   {%- set tmp_identifier = model['name'] + '__dbt_tmp' -%}
   {%- set backup_identifier = model['name'] + '__dbt_backup' -%}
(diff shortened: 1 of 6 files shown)


dbt/include/clickhouse/macros/materializations/table.sql
```diff
@@ -0,0 +1,85 @@
+{% materialization table, default %}
```
genzgd, line 1: Should this be `adapter='clickhouse'` instead of `default` just for consistency?

dbt/include/clickhouse/macros/materializations/view.sql
```diff
@@ -0,0 +1,89 @@
+{%- materialization view, default -%}
```
genzgd, line 1: Same, should this be `adapter='clickhouse'` instead of `default`

genzgd (approved):
Thanks for submitting this PR!  Atomic exchange is definitely an improvement.  It would be nice bonus if you could update the tests to cover both situations but I can work on adding that as time permits.

genzgd:
@Barsoomx   Do want us to do a 1.1.x release with these changes that you can use before we push a 1.2.x version?

genzgd:
As soon as I have a chance to do some additional validation.

genzgd:
1.1.8 tagged and released

PR #119, opened 2022-12-12
In apply grants use role_name when user_name not available

saurabhbikram:
Currently if the grant is given to a role, the `user_name` is null which leads to the below error when running `apply_grants`

```
'NoneType' object has no attribute 'casefold'
```


genzgd:
Thanks for your PR!  I'm wondering how this will work with the "apply grants" macros for materializations where the original relation includes role based grants.  The current dbt [test](https://github.com/dbt-labs/dbt-core/blob/df93858b4b1752950669f7e32cee82bccaf06263/tests/adapter/dbt/tests/adapter/grants/test_model_grants.py#L82-L152) for apply grants only uses "test users", not roles.  Have you manually tested that case?

saurabhbikram:
The apply grant/revoke syntax is the same for users and roles as Clickhouse does not care if we are giving a grant to a user or a role. https://clickhouse.com/docs/en/sql-reference/statements/grant/#granting-privilege-syntax

I have tested it manually and the materialization fails if grantee is a role without this hotfix. (you have to run the model a second time)

genzgd:
Sounds good, thanks!

PR #122, opened 2022-12-23
Release 1 3 2

genzgd:
New delete+insert strategy, s3source macro


genzgd:
Note the cloud test is failing because of a known issue with updating table comments.

PR #139, opened 2023-02-07
Sync replica before exchange on `exchange_tables_atomic`

saurabhbikram:
## Summary

Sync tables across replicas before exchange to preserve consistency. This was recommended to us by Altinity, we used it on our cluster without any degradation in performance.

### DIFF
--- a/dbt/include/clickhouse/macros/adapters.sql
+++ b/dbt/include/clickhouse/macros/adapters.sql
@@ -109,6 +109,11 @@
 {% endmacro %}
 
 {% macro exchange_tables_atomic(old_relation, target_relation, obj_types='TABLES') %}
+
+  {%- if adapter.get_clickhouse_cluster_name() is not none and structure == 'TABLES' %}
+    {% do run_query("SYSTEM SYNC REPLICA "+target_relation.identifier) %}
+  {%- endif %}
+  
   {%- call statement('exchange_tables_atomic') -%}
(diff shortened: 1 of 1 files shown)


dbt/include/clickhouse/macros/adapters.sql
```diff
@@ -109,6 +109,11 @@
 {% endmacro %}
 
 {% macro exchange_tables_atomic(old_relation, target_relation, obj_types='TABLES') %}
+
+  {%- if adapter.get_clickhouse_cluster_name() is not none and obj_types == 'TABLES' %}
+    {% do run_query("SYSTEM SYNC REPLICA "+target_relation.identifier) %}
```
simpl1g, line 114: @genzgd @saurabhbikram 
Don't we need to run it on cluster?
https://clickhouse.com/docs/en/sql-reference/statements/system/#sync-replica

I guess it will run sync only on single shard

genzgd: Yeah I think you're right.  I'm not very familiar with that command but syncing on cluster seems safer.

saurabhbikram: Yes we should definitely add it. Thanks for highlighting.

genzgd: I'll stick it in the new release

Issue #144, opened 2023-03-21
Missing identifier quoting


apisarenco:
### Describe the bug
Identifiers are not quoted even when the setting is on

### Steps to reproduce
1. Have as source databases that contain non-alphanumeric characters (ex. UUID as database name)
2. Have queries that source from those tables
3. Compile the model and behold! not quoted strings.

### Expected behaviour
Database names and table names should be quoted when configured

### Code examples, such as models or profile settings

```
version: 2

sources:
  - name: t01
    schema: "a6bf4b3f-17ee-4ae3-b769-acc2ffe0d708"
    tables: &tables
      - name: the_table
      - name: the_other_table
    quoting: &quoting
      database: true
      schema: true
      identifier: true

  - name: t02
    schema: "6410659e-85c7-413b-b952-aca7fd639788"
    tables: *tables
    quoting: *quoting
```

```
{{
    config({
        "materialized": "table",
        "engine": "MergeTree()",
        "schema": "dev",
        "order_by": '(id)'
    })
}}
{% set tenants =  ['t01', 't02'] %}
{% set data_sources = [] %}
{% for tenant in tenants %}
{% do data_sources.append(source(tenant, 'the_table')) %}
{% endfor %}
WITH cte AS (
    {{ dbt_utils.union_relations(data_sources) }}
)
SELECT * FROM cte
```

PR #145, opened 2023-03-21
Specify ANSI SQL quote character

apisarenco:
Relation names might need to be quoted. ANSI SQL style is to use double quotes, even though ClickHouse also supports backticks. Keep it consistent with the base class and use double quotes. Keep the attribute for future use.

## Summary
https://github.com/ClickHouse/dbt-clickhouse/issues/144

Specify quoting character so that identifiers get quoted

## Checklist
- [v] A human-readable description of the changes was provided to include in CHANGELOG


genzgd:
If we use the double quote character, don't we have to escape certain characters? I know it's not standard, but how do you feel about the backtick which is kind of standard for ClickHouse and doesn't normally require escaping?

PR #148, opened 2023-04-05
Reduce number of MODIFY COMMENT queries count

simpl1g:
## Summary
Call MODIFY COMMENT only if we really have description for column
Squash all alters for comments into single command to reduce number of SQL calls

Given this model
```yml
  - name: table
    description: Parsed events
    columns:
      - name: date
        description: Date when error occurred
      - name: timestamp
        description: timestamp from Kafka metadata
      - name: app_id
      - name: user_id
      - name: platform_id
```
Previous behaviour

```sql
alter table db.table modify comment 'Parsed events'
alter table db.table comment column date 'Date when error occurred'
alter table db.table comment column timestamp 'timestamp from Kafka metadata'
alter table db.table comment column platform_id ''
alter table db.table comment column app_id ''
alter table db.table comment column user_id ''
```

New behaviour
```sql
alter table db.table modify comment 'Parsed events', comment column date 'Date when error occurred', comment column timestamp 'timestamp from Kafka metadata'
```

### DIFF
--- a/dbt/include/clickhouse/macros/catalog.sql
+++ b/dbt/include/clickhouse/macros/catalog.sql
@@ -5,11 +5,11 @@
       columns.database as table_schema,
       columns.table as table_name,
       if(tables.engine not in ('MaterializedView', 'View'), 'table', 'view') as table_type,
-      tables.comment as table_comment,
+      nullIf(tables.comment, '') as table_comment,
       columns.name as column_name,
       columns.position as column_index,
       columns.type as column_type,
-      columns.comment as column_comment,
+      nullIf(columns.comment, '') as column_comment,
       null as table_owner
     from system.columns as columns
     join system.tables as tables on tables.database = columns.database and tables.name = columns.table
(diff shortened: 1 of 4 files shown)


genzgd:
Thanks for your PR?  Do you happen to know if this is covered by any of the existing tests?

simpl1g:
> Thanks for your PR? Do you happen to know if this is covered by any of the existing tests?

@genzgd Currently there are no tests for `persist_docs` functionality. I will try to import them from `dbt-core`. Looks like we have uncovered cases with missing columns and quoting

simpl1g:
@genzgd There is https://github.com/dbt-labs/dbt-core/blob/main/tests/adapter/dbt/tests/adapter/persist_docs/test_persist_docs.py test in upcoming 1.5.0 release of dbt. This is not yet available for us in current 1.4.5. So I copied full code without modifications, after migrating to new dbt version it can be replaced with just

```py
class TestPersistDocs(BasePersistDocs):
    pass


class TestPersistDocsColumnMissing(BasePersistDocsColumnMissing):
    pass


class TestPersistDocsCommentOnQuotedColumn(BasePersistDocsCommentOnQuotedColumn):
    pass
```

genzgd:
Awesome, thanks for the additional code and tests!

genzgd:
If we use the double quote character, don't we have to escape certain characters?  I know it's not standard, but how do you feel about the backtick which is kind of standard for ClickHouse and doesn't normally require escaping?

PR #154, opened 2023-05-17
Fix database creation when cluster name is defined

codenation-nl:
## Summary
When running an on-premise ClickHouse cluster, dbt-clickhouse will attempt to create the database but doesn't yet take cluster mode into account. With this PR, I attempted to solve this issue by taking into account by checking if a cluster name is set in the credentials. If it is present, it adds it to the create database clause respecting the syntax from https://clickhouse.com/docs/en/sql-reference/statements/create/database.

I've tested it, but as the whole dbt-clickhouse repo doesn't have tests for cluster mode, I've not added integration tests here.

## Checklist

### DIFF
--- a/dbt/adapters/clickhouse/dbclient.py
+++ b/dbt/adapters/clickhouse/dbclient.py
@@ -66,11 +66,7 @@ class ChClientWrapper(ABC):
         self._client = self._create_client(credentials)
         check_exchange = credentials.check_exchange and not credentials.cluster_mode
         try:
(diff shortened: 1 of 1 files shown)


genzgd:
Thanks for your PR!  The change will make it into the next patch release, which we'll try to do in the next week or so even if there aren't additional contributions.

codenation-nl:
Awesome, thanks for the quick reply. This will help us out a lot 👍

Issue #161, opened 2023-06-12
Incompatible return type `ClickHouseConnectionManager.execute`


PrVrSs:
### Describe the bug
`run_hooks` method raise exception: `Attribute: 'str' object has no attribute '_message'`. 
https://github.com/dbt-labs/dbt-core/blob/v1.4.1/core/dbt/task/run.py#L366

### Steps to reproduce
1. Create macro
```
{% macro parity_str_udf %}

CREATE FUNCTION parity_str AS (n) -> if(n % 2, 'odd', 'even');

{% endmacro %)
```
2. dbt run

### Expected behaviour
No exception.
### Code examples, such as models or profile settings
`dbt_project.yml`

```
on-run-start:
- {{ parity_str_udf() }}
```
#### Environment
* dbt version: 1.4.x
* dbt-clickhouse version: 1.4.2

ClickHouseConnectionManager.execute must return: `Tuple[AdapterResponse, agate.Table]`
Now: `Tuple[str, agate.Table]`

genzgd:
Thanks for your report and PR -- this seems like it would have been broken for quite some time, do you why this issue is just showing up now?

PrVrSs:
I don't know. I recently started using dbt.

PR #162, opened 2023-06-12
Fix incompatible return type

PrVrSs:
## Summary

#161 

Change `ClickHouseConnectionManager.execute` return type
`Tuple[str, agate.Table]` -> `Tuple[AdapterResponse, agate.Table]`

examples:
- https://github.com/dbt-labs/dbt-redshift/blob/main/dbt/adapters/redshift/connections.py#L390
- https://github.com/dbt-labs/dbt-snowflake/blob/main/dbt/adapters/snowflake/connections.py#L447


genzgd:
Looks good, thanks again -- I guess people don't use hooks very often or we would have seen this previously.  We'll do a 1.4.x patch release in the next few days.

PR #163, opened 2023-06-19
Distributed table materialization

gladkikhtutu:
Add distributed table materialization and local/shard/node suffix config

PR #166, opened 2023-07-12
fix: incremental on cluster clause

Savid:
Missing `ON CLUSTER` clause when creating incremental temporary replicated table.

**Error;**
```bash
Code: 36. DB::Exception: Macro 'uuid' and empty arguments of ReplicatedMergeTree are supported only for ON CLUSTER queries with Atomic database engine. (BAD_ARGUMENTS) (version 23.6.2.18 (official build))
```

**Debug sql**
```sql
create table dbt.example__dbt_tmp as dbt.example__dbt_new_data
```

**Example config;**
```sql
{{ config(
    materialized="incremental",
    engine="ReplicatedMergeTree('/clickhouse/{installation}/{cluster}/tables/{shard}/{database}/{table}/{uuid}', '{replica}')",
    ...
) }}
```

## What this fix does

Changes
```sql
create table dbt.example__dbt_tmp as dbt.example__dbt_new_data
```

to

```sql
create table dbt.example__dbt_tmp ON CLUSTER '{cluster}' as dbt.example__dbt_new_data
```

If cluster has been set in config.

### DIFF
--- a/dbt/include/clickhouse/macros/materializations/incremental.sql
+++ b/dbt/include/clickhouse/macros/materializations/incremental.sql
@@ -146,7 +146,12 @@
     -- use the table just created in the previous step because we don't want to override any updated rows with
     -- old rows when we insert the old data
     {% call statement('main') %}
-       create table {{ intermediate_relation }} as {{ new_data_relation }}
+        create table {{ intermediate_relation }}
+        {% set active_cluster = adapter.get_clickhouse_cluster_name() %}
+        {%- if active_cluster is not none %}
+        ON CLUSTER {{ active_cluster }}
+        {% endif %}
+        as {{ new_data_relation }}
     {% endcall %}
 
     -- Insert all the existing rows into the new temporary table, ignoring any rows that have keys in the "new data"


genzgd:
What is the point of replicating a temporary table?  I think we would have to update all of the temporary table operations (including the cleanup) as ON CLUSTER and the behavior would have to be different with ReplicatedMergeTrees versus a regular MergeTrees.

Savid:
> What is the point of replicating a temporary table? I think we would have to update all of the temporary table operations (including the cleanup) as ON CLUSTER and the behavior would have to be different with ReplicatedMergeTrees versus a regular MergeTrees.

I might be misunderstanding the process but that table is exchanged and should be replicated?

1. create `example__dbt_new_data` (using the model config engine and `ON CLUSTER` clause)
1. insert into `example__dbt_new_data` from model (with incremental filters)
1. create `example__dbt_tmp` as `example__dbt_new_data` (**this current fails as its missing the `ON CLUSTER` clause**)
1. insert into `example__dbt_tmp` from `example` excluding rows already in `example__dbt_new_data`
1. insert into `example__dbt_tmp` from `example__dbt_new_data`
1. drop `example__dbt_new_data` and `example__dbt_backup` if exists
1. rename `example__dbt_tmp` to `example__dbt_backup`
1. **exchange** `example__dbt_backup` to `example`

genzgd:
Sorry about that, you're correct, this particular table is not actually "temporary".

However, all of the other operations involved are not `ON CLUSTER`, so this code really can only work with a ReplicatedMergeTree engine.  Have you confirmed that incremental materializations work correctly `ON CLUSTER` with a ReplicatedMergeTree table?  Maybe we should add a check that the the model uses a Replicated Engine (not sure how that would work).  At a minimum we should probably explicitly document that incremental materializations do not work if a cluster is defined and the underlying model is not a *ReplicatedMergeTree table.

I think I'm okay with merging this if you've validated that the whole incremental materialization process works for ReplicatedMergeTrees (although automated tests would be really nice) and we have at least documented that ON CLUSTER must be combined with *ReplicatedMergeTrees for incremental materialization.

(Finally just as an FYI it looks like @gladkikhtutu is working on the related problem of incremental materialization of Distributed Tables - which also will involve `ON CLUSTER` operations. https://github.com/ClickHouse/dbt-clickhouse/pull/163)

gladkikhtutu:
I can add that we already use on cluster clause for intermediate table, for example [here](https://github.com/ClickHouse/dbt-clickhouse/blob/67cf9db5285b9d1c41ce321bd902ece577b6cc34/dbt/include/clickhouse/macros/materializations/incremental.sql#L40C3-L40C3) due to [this statement](https://github.com/ClickHouse/dbt-clickhouse/blob/67cf9db5285b9d1c41ce321bd902ece577b6cc34/dbt/include/clickhouse/macros/materializations/table.sql#L157). I think it works not only for Replicated table engines, but if you have multiple shards.
And yes, I will rewrite and reuse this place in distributed incremental materialization, I need to have on cluster clause and distributed table.

Savid:
I'm happy to close this and use my own fork for now. I can wait for distributed incremental materialization changes.

Thanks all!

Issue #167, opened 2023-07-17
SYSTEM SYNC REPLICA call in exchange_tables_atomic macro missing space


damirbk:
### Describe the bug

The fix implemented in #156 will format the SQL in the following way:
```sql
SYSTEM SYNC REPLICA

ON CLUSTER "{cluster}"db.table_name
```

As you can see, compared to [the documentation](https://clickhouse.com/docs/en/sql-reference/statements/system#sync-replica), there is a space missing between the ON CLUSTER clause and the db.table_name.

### Expected behaviour

The rendered SQL should look like this:
```sql
SYSTEM SYNC REPLICA

ON CLUSTER "{cluster}" db.table_name
```
So simply a space should be added.

#### Environment
* dbt-clickhouse version: 1.4.2

genzgd:
PR with fix has been merged and released in 1.4.4

PR #168, opened 2023-07-17
Add space to exchange_tables_atomic macro

damirbk:
This changes the SYSTEM SYNC REPLICA query to have a space between the ON CLUSTER clause and the table name.

Solves #167


genzgd:
Thanks for the fix!  Maybe we could just put that space in the on_cluster macro instead so we don't have the issue in the future?

damirbk:
Sure, that sounds like a good idea!

damirbk:
Struggled a bit with leaving the final space but still keeping the formatting the same otherwise (due to the leading `{%-` in the endif) so @genzgd please shout if you think the solution should be improved!

genzgd:
LGTM -- I agree it's tricky to figure out how to keep white space like that in the Jinja macros.

PR #172, opened 2023-07-21
Distributed incremental materialization

gladkikhtutu:
Created distributed incremental materialization with docs.

To add tests, clickhouse with cluster should be added in docker compose.


gfunc:
IMO it is crucial to enable `insert_distributed_sync` for distributed table inserts for dbt usage ( the default is disabled for ClickHouse 22.12). Otherwise, downstream calculation could be wrong if the distributed insert is not finished in time.

gladkikhtutu:
> IMO it is crucial to enable `insert_distributed_sync` in ClickHouse for distributed table inserts for dbt usage ( the default is disabled for ClickHouse 22.12). Otherwise, downstream calculation could be wrong if the distributed insert is not finished in time. Maybe add a check `SELECT value FROM system.settings WHERE name = 'insert_distributed_sync'; ` as prehook, or add `SET insert_distributed_sync = 1` before inserting SQL run, or a notice in README?

Yes, I added warning about `insert_distributed_sync` to README and exception in materialization if `insert_distributed_sync` != 1. Don't know, if `SET insert_distributed_sync = 1 before inserting SQL` is a correct solution, maybe should think about it in seperate PR.

README.md
```diff
@@ -153,6 +156,106 @@ keys used to populate the parameters of the S3 table function:
...
+
+Only shard tables are replacing, because distributed table does not keep data. 
+The distributed table reloads only when the full_refresh mode is enabled or the table structure may have changed.
+
+### Distributed incremental model example
+```sql
+{{
+    config(
+        materialized='distributed_incremental',
+        on_cluster='cluster',
```
genzgd, line 227: Should it just use the `on_cluster_clause` macro instead of specifying it as a parameter?  I think every other operation will require that macro to be automatically applied using the "active" cluster, so if the wrong cluster is specified here, that would break stuff?

genzgd: Actually it looks like that's what you do, so we just need to remove it from the example here?

dbt/include/clickhouse/macros/materializations/distributed_table.sql
```diff
@@ -98,3 +107,13 @@
   {{ partition_cols(label="partition by") }}
   {{ adapter.get_model_settings(model) }}
 {%- endmacro %}
+
+{% macro create_distributed_shard_table(distributed_relation, shard_relation, structure_relation, sql_query=none) -%}
```
genzgd, line 111: nit -- maybe we should call this distributed_local table to match the local prefix?

genzgd (changes requested):
Thanks for the hard work here!  My comments are generally minor and before releasing I will probably add some notes to the documentation that this is "experimental" and not currently part of our test suite, but I think it's pretty close.

README.md
gladkikhtutu: Yes, removed

PR #186, opened 2023-08-23
Cluster setting and Distributed Table tests

gfunc:
## Summary
- added attribute should_on_cluster for ClickHouseRelation class to indicate whether the`on_cluster_clause` macro should return content.
- added cluster setting for tests.
- debug marcos.
- added tests for distributed_table materialization and replicated table engines.
- changed github workflow test to use docker-compose which starts a 3 nodes ClickHouse cluster

## Checklist

## Caveats
- ~~tests for distributed_incremental is not added yet.~~


zli06160 (commented):
👍 The modifications in `dbt/include/clickhouse/macros/materializations/distributed_table.sql` do make sense. 
We should indeed 

- define the default `sharding_key`, `cluster`, etc;
- or show the error messages(instructions) if they are not defined in `template.sql` and/or `dbt_project.yml`.

gfunc:
any update on the review of this MR?

zli06160:
@gfunc Sorry I will not be able to contribute this week, not sure about the next week.

tests/integration/adapter/incremental/test_distributed_incremental.py
```diff
@@ -0,0 +1,205 @@
...
+replicated_seed_schema_yml = """
+version: 2
+
+seeds:
+  - name: base
+    config:
+      engine: ReplicatedMergeTree('/clickhouse/tables/{uuid}/one_shard', '{server_index}' )
+  - name: added
+    config:
+      engine: ReplicatedMergeTree('/clickhouse/tables/{uuid}/one_shard', '{server_index}' )
```
zli06160, line 153: Maybe it will be better if test at least once with _**multiple shards** x **multiple replicas**_?
e.g. `ReplicatedMergeTree('/clickhouse/tables/{shard}/{database}/table_name', '{replica}')` (example from https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication)

zli06160:
@gfunc  I think you forgot to rebase on the up-to-date main branch. 
(The changes in https://github.com/ClickHouse/dbt-clickhouse/pull/184/files are currently missing)
The `distributed_table` works well after the rebase/sync.

zli06160:
The `distributed_table` works well on the cluster with my test case.

gfunc:
Hi @zli06160 any thoughts on the logic for should_on_cluster attribute in relation.py? 

```python
    @property
    def should_on_cluster(self) -> bool:
        if self.include_policy.identifier:
            return self.can_on_cluster
        else:
            # create database/schema on cluster by default
            return True
    @classmethod
    def get_on_cluster(
        cls: Type[Self], cluster: str = '', materialized: str = '', engine: str = ''
    ) -> bool:
        if cluster:
            return 'view' == materialized or 'distributed' in materialized or 'Replicated' in engine
        else:
            return False
```

I think we should start the discussion early since this is changing the behavior of this adapter once a cluster is set. Also, I tried to make logic compatible with the previous version by adding the attribute `can_on_cluster` for existing models. 

Maybe it is worth mentioning in README?

zli06160:
The logic of `get_on_cluster` seems correct, just need to include [`MATERIALIZED VIEW`](https://clickhouse.com/docs/en/sql-reference/statements/create/view) (of course this materialization is not implemented yet), and prevent the case when `cluster = '  '`.

gfunc:
I will be available later starting tomorrow (Oct 09) for further modifications.

tests/integration/adapter/incremental/test_distributed_incremental.py
```diff
@@ -0,0 +1,205 @@
...
+    model_incremental,
+    schema_base_yml,
+    seeds_added_csv,
+    seeds_base_csv,
+)
+from dbt.tests.adapter.basic.test_incremental import BaseIncremental, BaseIncrementalNotSchemaChange
+from dbt.tests.util import run_dbt
+
+from tests.integration.adapter.incremental.test_incremental import uniq_schema
+
```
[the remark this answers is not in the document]
gfunc: I am afraid most tests are self explainatory by names, and mostly base off dbt's sample tests. I will try to add more comments for distributed materialization tests.

tests/integration/adapter/incremental/test_distributed_incremental.py
gfunc: for integration tests, if we managed to set a multiple shards * multple replicas matrix, data will always exist in one shard (same zk path), maybe it is not worth adding another test?

zli06160: For me it is an usual usage.
But it is more managed by **dbt-core** & the **clickhouse cluster**, so any integration test on the _cluster_ mode is enough.

zli06160: BTW, I did not get the reason why the data persist in one shard.
Do you use `drop table distributed_table_name on cluster cluster_name` or `create or replace table`, or something similar?

gfunc: I meant for the Replicated Engine tables, not distributed tables. If I understand it correctly, during our tests, `insert into` query will be performed by one node and data will be replicated among nodes with the same shard (same zk path).

gfunc: I am not sure if I understand you correctly. 
`Seeds` in test `TestInsertsOnlyDistributedIncrementalMaterialization` seem to serve the purpose of providing source data and seed function is already being tested in `test_basic.py` but with table materialization only. 

My quetstion is, should we add another test for seed with distributed materialization and check data distribution or we can add another test like `TestDistributedMaterializations` in `test_basic.py` with a different cluster (maybe 1*2 + 1*1) setting?

gfunc: My approach is to make sure seed are using the same shard and different replica, i.e. to broadcast the whole dataset.
I have never used File Table Engine before, could you explain a bit more? How it would help in this situation?

zli06160: `File Table Engine` reads the csv files directly, not sure if it will help.
https://clickhouse.com/docs/en/engines/table-engines/special/file

tests/integration/adapter/test_basic.py
```diff
@@ -9,7 +11,13 @@
 from dbt.tests.adapter.basic.test_singular_tests import BaseSingularTests
 from dbt.tests.adapter.basic.test_snapshot_check_cols import BaseSnapshotCheckCols
```
zli06160, line 12: I think it is better to set `"DBT_CH_TEST_CLUSTER"`(I guess the value should be 'test_shard'?) somewhere. After I cloned the project, build + up the docker containers, (without any manuel modifications) there is no such variable. 

To fully run the tests, with
 `os.environ["DBT_CH_TEST_CLUSTER"] = 'test_shard'` added, 
there is an error related to `BaseSnapshotCheckCols`:
```bash
09:11:29  Database Error in snapshot cc_all_snapshot (snapshots/cc_all_snapshot.sql)
09:11:29    :HTTPDriver for http://localhost:8123 returned response code 400)
09:11:29     Code: 15. DB::Exception: Column dbt_change_type specified more than once. (DUPLICATE_COLUMN) (version 23.9.1.1854 (official build))
```

tests/integration/adapter/incremental/test_distributed_incremental.py
zli06160: I think now I understand how you test the `distributed_table`. I still recommend a separate `.py` for each main materialization, or, if not applicable, document more.

tests/integration/adapter/test_basic.py
gfunc: Yes, I am using `test_shard` for unit tests. And I added the env var in [github workflow file](https://github.com/gfunc/dbt-clickhouse/blob/4e592222e12fa40aab35241bccf7c51466662782/.github/workflows/test_matrix.yml#L62). 
And tests returned no error. I will take a look at this test.

gfunc: I am not seeing any problem with this test.env file
```env
DBT_CH_TEST_USE_DOCKER=true
DBT_MACRO_DEBUGGING=true
DBT_CH_TEST_INCLUDE_S3=false
DBT_CH_TEST_CLUSTER=test_shard
```
and command
```bash
pytest tests/integration/adapter/test_basic.py::TestSnapshotCheckCols
````

result attached:
```
============================= test session starts ==============================
platform linux -- Python 3.9.18, pytest-7.4.0, pluggy-1.2.0
rootdir: /home/xxx/projects/dbt-clickhouse
configfile: pytest.ini
plugins: dotenv-0.5.2
collected 1 item

tests/integration/adapter/test_basic.py .                                [100%]

============================== 1 passed in 19.58s ==============================

```

Please share your env file for me to reproduce

Comment (zli06160) [reply tests/integration/adapter/test_basic.py:12]: The reason is that I didn't have the `test.env`.

With the same env file as yours, 
```bash
pytest tests/integration/adapter/test_basic.py -c pytest.ini -s
```
=> everything is ok after double check test by test.

```bash
pytest tests/integration/ -c pytest.ini
```
=> 107 passed, 2 skipped, 2 warnings .

zli06160 (approved):
OK for me 👍

genzgd:
Thanks for all the work here, @gfunc and @zli06160.  I'll do a 1.4.9 release shortly.

Issue #128, opened 2023-01-08
Support insert_overwrite Materialization Strategy


genzgd:
We may want to consider also implementing an `insert+overwrite` incremental materialization strategy for models with many ClickHouse partitions.  This would involve dropping partitions that have new or updated data and doing a complete replacement of such partitions (possibly from a temporary table.)

To make this performant we could have to analyze the number and size of the partition affected by the update and weigh delete+insert cost against the cost of simply dropping partitions and replacing the data.

The other consideration might be how strategies interact with ClickHouse transactions when and if production transactional support is available.

simpl1g:
Sounds similar to what we have now internally

We introduced extra option in model config

```python
{{
  config (
      ...
      partition_by = 'toYYYYMM(date)',
      partition_strategy = 'append', // or can be 'rewrite'
  )
}}
```

- create tmp table
- insert new data in tmp table
- get all partitions in tmp table
- if partition_strategy = 'append' => ATTACH PARTITION ... FROM tmp
- if partition_strategy = 'rewrite' => REPLACE PARTITION ... FROM tmp
- drop tmp

You just need to be careful when doing replace, to not lose anything :) For example if you have monthly partitioning you need to make sure that you update whole month, not only last X days

bryzgaloff:
Hi @simpl1g thank you for sharing, your approach sounds the best balance of fault-tolerance and efficiency. Could you please share an implementation with the community? I will be glad to review the contribution and give my approval to speed up the release.

Without this feature ClickHouse DBT plugin looks quite unusable for production data pipelines due to complex recovery from failures. I am looking for options to integrate DBT with Airflow to manage ClickHouse partitions based on Airflow DAG's execution date (e.g. partition tables daily and then update them via daily Airflow DAG using insert-replace strategy).

PR #201, opened 2023-10-26
insert_overwrite (insert+replace) partitions incremental strategy

bryzgaloff:
## Summary
This PR implements insert+replace strategy discussed in #128 which does the following:
* Creates a new staging table with the same structure as the target table.
* Inserts data into the staging table.
* Replaces partitions in the target table from the staging table.

Advantages:
* Only the involved partitions are replaced: this is much cheaper than reinserting the full table which is implemented in other strategies.
* If an insertion fails, the target table is not affected.

## Checklist


bryzgaloff:
At the time of publishing the PR is in WIP (work in progress) state since I need an advice from the community. Thus, neither documentation nor tests are updated.

dbt/include/clickhouse/macros/materializations/incremental.sql
```diff
@@ -234,3 +244,36 @@
...
+          from {{ intermediate_relation }}
+      {% endset %}
+      {% set partitions_expressions = run_query(select_changed_partitions).rows %}
+    {% else %}
+      {% set partitions_expressions = [] %}
+    {% endif %}
+    {% for partition_expression in partitions_expressions %}
+      {% call statement('replace_partition_' ~ loop.index) %}
+          alter table {{ existing_relation }}
+          replace partition {{ partition_expression }}
```
bryzgaloff, line 273: I am new to DBT and I need an advice for translating output of `run_query` into ClickHouse literal values. Currently `partitions_expressions` is an iterable of [`agate.Row`s](https://agate.readthedocs.io/en/latest/api/columns_and_rows.html#agate.Row). I need to translate `Row{someInt: 1, someStr: 'a', someDate: toDate('2023-11-01')}` to partition expression like `… partition (1, 'a', toDate('2023-11-01')`.

Additionally, we have `partition_by` list of columns names (without types) as an input.

I have several ideas to do it:
* Request target table's columns definitions. Pick the partitioning columns types. Use `CAST('<value-as-string-literal>' AS <target-type>)`.
* Implement translation as a method of `ClickHouseAdapter` and call `adapter.to_partition_expression(…)`. Though I am not sure about the best way to perform translation in adapter: I hope to leverage [`clickhouse-driver`](https://clickhouse-driver.readthedocs.io/en/latest/api.html#clickhouse_driver.Client.execute)'s functionality of translating `params` for this. However, the driver might not have a convenient public interface for the translation.
* Find out some SQL-only way to translate. I.e. compose an SQL query which will do the translation on ClickHouse level: smth like `with partition_expr as (select partitionBy offset {{ loop.index }} limit 1) alter table … replace partition partition_expr …`. This particular example does not work, but maybe there is a possible way.
* I am not sure if `clickhouse-driver`'s usage is discouraged in favour of HTTP interface. If so, a parameterized HTTP query might be an option, though looks the most complex option to me and still requires retrieving partitioning columns' types.

Please let me know if there is an easier dbt-style way to do it. Any advice regarding the suggested options or guidance towards using dbt-specific functions (like requesting a list of table's columns definitions) will be helpful.

I am doing my investigation of the approaches to choose the simplest one. Though I believe the first option is the most reasonable.

@genzgd @simpl1g @tanghyd given that you participated in the original issue #128, could you maybe share some thoughts please?

dbt/include/clickhouse/macros/materializations/incremental.sql
```diff
@@ -41,10 +41,14 @@
     {% endcall %}
     {% set need_swap = true %}
 
-  {% elif inserts_only or unique_key is none -%}
+  {% elif
+      inserts_only
+      or unique_key is none
+      and config.get('incremental_strategy', none) != 'insert+replace' -%}
```
bryzgaloff, line 47: A mix of legacy (`insert_only`) and new (`"append"` strategy) approaches to configure the strategy introduces counter-intuitive set of configuration in dbt-clickhouse. To not break the current conditions check flow I have implemented this dirty way of checking, though in general I would suggest refactoring these conditions tree.

To narrow down the reviewed scope, I won't blend the changes into this PR, thus I suggest keeping this dirty-yet-working workaround here.

dbt/include/clickhouse/macros/materializations/incremental.sql
bryzgaloff, line 273: Also, I have doubts about replacing partitions here in a loop: when multiple partitions are affected, this makes the operation non-atomic: if a subset of partitions is not replaced, the target table might end up being in an inconsistent state. Does dbt have a rollback functionality?

Or does ClickHouse supports manipulating multiple partitions at once? I have not found in documentation and could not compose such a query. Though many Big Data tools (e.g. AWS Athena, Hive) support multiple partitions manipulation when the partitions are listed in a comma-separated fashion.

dbt/adapters/clickhouse/impl.py
```diff
@@ -113,7 +113,7 @@ def calculate_incremental_strategy(self, strategy: str) -> str:
         if not strategy or strategy == 'default':
             strategy = 'delete_insert' if conn.handle.use_lw_deletes else 'legacy'
         strategy = strategy.replace('+', '_')
-        if strategy not in ['legacy', 'append', 'delete_insert']:
+        if strategy not in ['legacy', 'append', 'delete_insert', 'insert_replace']:
```
simpl1g, line 116: I believe `insert_overwrite` is more idiomatic name for this strategy

dbt/include/clickhouse/macros/materializations/incremental.sql
```diff
@@ -72,6 +76,12 @@
       {% call statement('main') %}
         {{ clickhouse__insert_into(target_relation, sql) }}
       {% endcall %}
+    {% elif incremental_strategy == 'insert_replace' %}#}
+      {%- set partition_by = config.get('partition_by') -%}
```
simpl1g, line 80: Maybe also better to raise exception if `partition_by` is empty

dbt/include/clickhouse/macros/materializations/incremental.sql
```diff
@@ -234,3 +244,36 @@
     {% do adapter.drop_relation(new_data_relation) %}
     {{ drop_relation_if_exists(distributed_new_data_relation) }}
 {% endmacro %}
+
+{% macro clickhouse__incremental_insert_replace(existing_relation, intermediate_relation, partition_by) %}
```
simpl1g, line 248: this strategy should also take care about cluster setup and distributed tables. 
We can insert data through Distributed table and then do replace for local table on each shard. This part will be mostly the same as in `clickhouse__incremental_delete_insert` macro

dbt/include/clickhouse/macros/materializations/incremental.sql
```diff
@@ -234,3 +244,36 @@
...
+    {%- endcall %}
+    {% call statement('main') -%}
+        create table {{ intermediate_relation }} as {{ existing_relation }}
+    {%- endcall %}
+    {% call statement('insert_new_data') -%}
+        insert into {{ intermediate_relation }} select * from {{ new_data_relation }}
+    {%- endcall %}
+    {% if execute %}
+      {% set select_changed_partitions %}
+          select distinct {{ partition_by|join(', ') }}
```
simpl1g, line 263: we can do something like this to get partitions
this way we are guaranteed to select all partitions in temp relation
and and will be faster
```sql
{% set partitions = get_partitions(relation) %}
...
{% macro get_partitions(relation) %}
  {% set cluster = adapter.get_clickhouse_cluster_name() %}
  {% set source = 'system.parts' %}
  {% if cluster is not none %}
    {% set source = "cluster('{{ cluster }}', system.parts)" %}
  {% set sql -%}
    SELECT DISTINCT partition_id
    FROM {{ source }} WHERE active AND database = '{{ relation.schema }}' AND table = '{{ relation.identifier }}'
  {%- endset -%}

  {{ return(run_query(sql)) }}
{% endmacro %}

```

dbt/include/clickhouse/macros/materializations/incremental.sql
simpl1g: You can actually do it like this
As an example how to do it in jinja you can [look here](https://github.com/ClickHouse/dbt-clickhouse/blob/b79669ae555f5b060ef42dc14a3d6372cd418b33/dbt/include/clickhouse/macros/persist_docs.sql#L13), it does the same with comments array
```sql
ALTER TABLE ... REPLACE PARTITION 1, REPLACE PARTITION 2, REPLACE PARTITION 3
```

But if you have cluster it will not be atomical. One shard can work and other will fail.

simpl1g: See `get_partitions` macro in comment above, I believe it should solve this question

dbt/adapters/clickhouse/impl.py
bryzgaloff: Is it common in dbt plugins domain? I believe insert+replace is more ClickHouse-specific since it describes what the partition does exactly: "**inserts** into a staging table, then **replaces** partitions into a target table"

If it is smth more dbt-like, I agree to rename

dbt/include/clickhouse/macros/materializations/incremental.sql
bryzgaloff: Does not `.get` raise an exception in Jinja if the key is missing? Or does it have regular Python's dict behaviour?

dbt/include/clickhouse/macros/materializations/incremental.sql
bryzgaloff: Very helpful, thank you for your valuable feedback, Konstantin! 🤝 

I have implemented and tested you suggestion with selecting from `system.parts` and it works perfectly fine 👍

dbt/include/clickhouse/macros/materializations/incremental.sql
bryzgaloff: Should we maybe defer the implementation till the next iteration? My immediate requirement was to support the strategy for single-node setups (which we are using currently with my team). We may leave a chance to others implement and test the cluster-specific version. I won't have a quick option to test it, thus I won't be able to confirm it works.

Your suggestion with `system.parts` works completely fine, thanks for that! I plan to cover it with tests and document in the next couple of weeks (having a vacation next week :) ).

dbt/adapters/clickhouse/impl.py
simpl1g: It is more dbt specific, for example bigquery or databricks adapters implement it in similar way. You can check example here: https://docs.getdbt.com/reference/resource-configs/bigquery-configs#the-insert_overwrite-strategy

Also this is a good example for README, you can use it to update dbt-clickhouse documentation

dbt/include/clickhouse/macros/materializations/incremental.sql
simpl1g: Not sure about merging without Distributed support, because this is new functionality and mostly experimental, I think this is a question to @genzgd, as he is a maintainer.

But I believe that it should work with cluster from beginning because this is core CH functionality, and most of the production users don't use single server setup

dbt/adapters/clickhouse/impl.py
bryzgaloff: Hi Konstantin, I have reviewed other plugins and I still insist on `insert+replace` naming due to its unambiguity: the strategy name expresses exactly what it does. And this is an advantage of a suggested naming to me.

However, I rely on the package maintainers' decision here, so please let me know if you force me to change it.

dbt/include/clickhouse/macros/materializations/incremental.sql
bryzgaloff: We use dbt with ClickHouse without cluster currently, so I have not had an opportunity to test it. I suggest on merging the tested version of my PR now. Once someone needs a clustered version of the strategy, they may contribute it too and test themselves.

If I add the clusters-relates logic, I may not guarantee that it works. However, if the maintainers are ok to rely on their review, I am completely fine with adding a related code snippet.

bryzgaloff:
Hi @simpl1g and @genzgd, I've wrapped up this PR for `insert+replace` strategy implementation: added some fancy documentation in the README and threw in integration tests (taking a page from lw-deletes). Motivation for the feature is given in the PR's description.

Some discussions above are unresolved, waiting on your pearls of wisdom there :)

Please take a peek at the PR when you get a chance, and hit me up with your thoughts. Hoping we can merge the feature soon and let the good times roll with the new strategy! :)

bryzgaloff:
Hi @simpl1g and @genzgd! This is a kind reminder about my PR which is ready for your review. **We have successfully battle-tested it internally** ✅ 

We install my version from GitHub currently. It would be nice if you can approve it and release to PyPI. If any adjustments are required, please let me know! 🙏

genzgd:
@bryzgaloff I apologize that we haven't yet had the resources to fully review this PR.  As you may have noticed we've been focused on bug fixes and compatibility with the new dbt releases.  Please know that we very much appreciate the contribution (especially with test cases and real world usage) and your work is next on the roadmap as we get time.

If you have a chance to resolve the conflicts over the next few weeks that would be appreciated and make the review just a bit easier.

dbt/include/clickhouse/macros/materializations/incremental.sql
```diff
@@ -234,3 +247,42 @@
...
+       + '__dbt_new_data_' + invocation_id.replace('-', '_')}) %}
+    {{ drop_relation_if_exists(new_data_relation) }}
+    {% call statement('create_new_data_temp') -%}
+      {{ get_create_table_as_sql(False, new_data_relation, sql) }}
+    {%- endcall %}
+    {% call statement('main') -%}
+        create table {{ intermediate_relation }} as {{ existing_relation }}
+    {%- endcall %}
+    {% call statement('insert_new_data') -%}
+        insert into {{ intermediate_relation }} select * from {{ new_data_relation }}
```
dev-mkc19, line 262: @bryzgaloff as you said you has a single node deployment.  Here is should be  **{{ on_cluster_clause()}}**. Otherwise error below occurs in my deployment. I'm currently working on adaptation to work on cluster.
```
15:04:39    :HTTPDriver for http://10.100.0.106:8123 returned response code 400)
15:04:39     Code: 36. DB::Exception: Macro 'uuid' and empty arguments of ReplicatedMergeTree are supported only for ON CLUSTER queries with Atomic database engine. (BAD_ARGUMENTS) (version 24.2.1.2248 (official build))
```

Also I have to configure model adding allow_nullable_key option because of otherwise another error occurs. Did you meet with that?
```
14:18:56    :HTTPDriver for http://10.100.0.106:8123 returned response code 400)
14:18:56     Code: 44. DB::Exception: There was an error on [10.100.0.106:9000]: Code: 44. DB::Exception: Partition key contains nullable columns, but merge tree setting `allow_nullable_key` is disabled. (ILLEGAL_COLUMN) (version 24.2.1.2248 (official build)). (ILLEGAL_COLUMN) (version 24.2.1.2248 (official build))
```
```
{{
config(
    materialized = "incremental",
    partition_by = "transaction_date_part",
    incremental_strategy = "insert+replace",
    engine = 'ReplicatedMergeTree',
    order_by = ['city_id', 'transaction_date_part'],
    schema = "analytics",
    settings = {'allow_nullable_key': 1}
)
}}
```

dbt/adapters/clickhouse/impl.py
BentsiLeviav: Hi @bryzgaloff 

Thanks for your contribution!
I tend to agree with @simpl1g, as this is the semantic used with dbt (you can read about it [here](https://docs.getdbt.com/docs/build/incremental-strategy)).

I think it is important to align our adapter with the community standards (we can emphasize and detail how this was implemented in the readme file)

BentsiLeviav:
@bryzgaloff 
Thanks again for your contribution!
I would like to review your PR and merge it within the next few days. Can you sync your fork with the main repo? We upgraded dbt-core to 1.8.0. 

After syncing I'll review this one right away.

dbt/include/clickhouse/macros/materializations/incremental.sql
bryzgaloff: Hi @dev-mkc19, thank you for your input! I plan to get back to working on the PR next week. For now, I will keep it without cluster support as I do not have quick-to-setup infrastructure to test it. Feel free to make your own PR adding cluster support 🤝 I may review it to ensure it does not break any of the `insert+replace` semantics. Tag me as a reviewer once it is published!

bryzgaloff:
Hi @BentsiLeviav thank you (and the other reviewers, of course!) for your participation. I am not actively using the plugin right now, but I may get back to handling your review feedback late next week. If the conflicts are not too critical, I might be able to resolve them quickly.

dbt/adapters/clickhouse/impl.py
bryzgaloff: Ok, updated to `insert_overwrite`, please review the latest version, guys! :)

bryzgaloff:
Hi @BentsiLeviav I have rebased my contribution onto the latest `main` of this repository. Also, I have renamed the strategy to `insert_replace`. Please have a look! 👋

bryzgaloff:
Does this repo has automated tests? I do not have a quick infra to test the contribution, but if there is not testing workflow, I will perform the manual testing.

bryzgaloff:
Hi @BentsiLeviav and @genzgd — will you have a chance review and merge the PR soon? I have updated everything according to the review feedback and all the checks have passed. I would like to avoid the need to rebased it again 😅

BentsiLeviav:
Hi @bryzgaloff 
Huge thanks for this.
I will review and update you within the next few days.

BentsiLeviav (approved):

bryzgaloff:
Thank you @BentsiLeviav for the approval! What are the next steps for the PR to be merged?

BentsiLeviav:
Hi @bryzgaloff 
Sure, thank you for your contribution!

Before merging this, could you please add to the doc that this feature is experimental, and wasn't tested with cluster setup? 
It is crucial to highlight these 2 points.

Once we are done with that, I'll merge your PR.

Thanks again for your work!

BentsiLeviav:
Never mind, I'll take care of it :)

pheepa:
@BentsiLeviav @genzgd 

Hi! I’m really interested in contributing to this project, particularly with the cluster functionality.

Before diving into development, I want to ensure I write robust tests for the support of ReplicatedMergeTree and Distributed.

Here’s my plan:

 • Add TestInsertReplaceIncremental (adapted for Replicated) to test_base_incremental.py.
 • Add another version of TestInsertReplaceIncremental (for materialized=distributed_incremental).

Does this approach align with the project’s practices, or would you recommend any adjustments?

Thanks in advance for your guidance!

pheepa:
[pr](https://github.com/ClickHouse/dbt-clickhouse/pull/394) here it is!

lddlww:
got it, thanks for reply and explanation clearly

PR #123, opened 2022-12-28
(feat): add support for matview

israelg99:
## Introduction
Supports a `target_table` (`TO table`) for Materialized View, or it'd use its own table, which is fully configurable: engine, order by, partition by, etc... defaults to `MergeTree() order by tuple()`.

## Usage
A model to define a target table:
```sql
{{ config(order_by='(updated_at, id, name)', engine='MergeTree()', materialized='incremental', unique_key='id', inserts_only=True) }}

with actor_summary_table as (
    select id,
        name,
        num_movies,
        avg_rank,
        genres,
        directors,
        updated_at
    from {{ ref('actor_summary_transform') }}
) select *
from actor_summary_table
```
A mode to define a materialized view:
```sql
{{ config(materialized='view', matview=True, target_table='actor_summary_table`) }}

with actor_summary_transform as (
    SELECT id,
        any(actor_name) as name,
        uniqExact(movie_id)    as num_movies,
        avg(rank)                as avg_rank,
        uniqExact(genre)         as genres,
        uniqExact(director_name) as directors,
        max(created_at) as updated_at
    FROM (
        SELECT {{ source('imdb', 'actors') }}.id as id,
            concat({{ source('imdb', 'actors') }}.first_name, ' ', {{ source('imdb', 'actors') }}.last_name) as actor_name,
            {{ source('imdb', 'movies') }}.id as movie_id,
            {{ source('imdb', 'movies') }}.rank as rank,
            genre,
            concat({{ source('imdb', 'directors') }}.first_name, ' ', {{ source('imdb', 'directors') }}.last_name) as director_name,
            created_at
    FROM {{ source('imdb', 'actors') }}
        JOIN {{ source('imdb', 'roles') }} ON {{ source('imdb', 'roles') }}.actor_id = {{ source('imdb', 'actors') }}.id
        LEFT OUTER JOIN {{ source('imdb', 'movies') }} ON {{ source('imdb', 'movies') }}.id = {{ source('imdb', 'roles') }}.movie_id
        LEFT OUTER JOIN {{ source('imdb', 'genres') }} ON {{ source('imdb', 'genres') }}.movie_id = {{ source('imdb', 'movies') }}.id
        LEFT OUTER JOIN {{ source('imdb', 'movie_directors') }} ON {{ source('imdb', 'movie_directors') }}.movie_id = {{ source('imdb', 'movies') }}.id
[...]

PR #207, opened 2023-11-13
Add materialization macro for materialized view

SoryRawyer:
## Summary
Change: Adds a materialization macro to create and update materialized views.

Unlike #123, this is a separate materialization than tables and views and will create both the target table and the materialized view from one query. The model name is used for the target table and the materialized view appends `_mv` to the model name.

Since this reuses functionality from the table materialization, the configuration properties (`order_by`, `primary_key`, etc.) should work as expected. Similarly, the cluster clause seems to be added when appropriate.

## Checklist

### DIFF
--- a/dbt/include/clickhouse/macros/materializations/materialized_view.sql
+++ b/dbt/include/clickhouse/macros/materializations/materialized_view.sql
@@ -1,0 +1,120 @@
+{#-
+  Create or update a materialized view in ClickHouse.
+  This involves creating both the materialized view itself and a
+  target table that the materialized view writes to.
+-#}
+{%- materialization materialized_view, adapter='clickhouse' -%}
+
+  {%- set target_relation = this.incorporate(type='table') -%}
+  {%- set mv_name = target_relation.name + '_mv' -%}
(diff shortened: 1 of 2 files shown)


SoryRawyer:
Hey @genzgd, sorry to be a bother, but I was just wondering if you had any thoughts on this PR. In addition to the tests included in this PR, I've been able to use this branch successfully in a project that uses materialized views. If you have doubts about these changes, is there anything that would increase your confidence in this feature?

genzgd:
My apologies for the delay.  I have been working on a 1.6 release and I wanted to understand how this PR fit in with the `dbt-core` support for MVs in 1.6.  I'm hoping to have a 1.6 release this week and I'm tentatively planning to merge your work into that as well, with some documentation/explanation of how "ClickHouse materialized views aren't the same as Postgres/Oracle."

The basic approach seems okay on a quick scan but I'll dig deeper into it over the next few days.

SoryRawyer:
Thanks for the quick reply! I didn't test against dbt-core 1.6 too much, but I did keep their MV implementation in mind with the goal of easy (if not seamless) integration. dbt should still be able to dispatch to this materialized view implementation, but if that's not the case then I can update this PR (or submit follow-up changes).

PR #222, opened 2023-12-13
Wrap columns_in_query query in select statement

ptemarvelde:
## Summary
Appending `LIMIT 0` to queries that already have a LIMIT results in a syntax error:
```
... limit 10000 LIMIT 0 FORMAT JSON ...
>>> Code: 62. DB::Exception: Syntax error: failed at position 6358 ('LIMIT') (line 200, col 17): LIMIT 0
 FORMAT JSON.
 ```
This is resolved by wrapping the query inside a select statement and limiting this clause.


genzgd:
Thanks for the fix!  (Sorry about the formatting stuff, it's overkill imo but we haven't loosened it up yet.)

PR #228, opened 2023-12-23
allows to add a comment in table's or view's metadata

### LINKED ISSUE #227: materialization comment in metadata
I just want to add the possibility to write the comment in materialization metadata in Clickhouse

Tested it and it can be done by adding a few lines of code, for example, in [view materialization](https://github.com/ClickHouse/dbt-clickhouse/blob/ca9da0b6ebacc14d67139b0b417e0d4083b68a9c/dbt/include/clickhouse/macros/materializations/view.sql#L81) right after `as ( {{ sql }} )` like that:
```
...
  as (
    {{ sql }}
  )
    {% set comment = config.get('comment') %}
    {% if comment %}
      COMMENT '{{ comment }}'
    {% else %}
      COMMENT 'DBT'
    {%- endif %}
...
```

And with the customised `comment` in schema.yml it will be in CH
```
version: 2
models:
  - name: MODEL_NAME
    config:
      comment: SOME CUSTOM COMMENT
```

I can add it by myself and request a merge, but I'm totally new to contributing to anything at all so will be glad if somebody can guide how to contribute correctly :)

dstsimokha:
## Summary

closes #227 
Now the `comment` from the config is added to Clickhouse materialization metadata, that's all (probably that's the smallest commit possible)


BentsiLeviav:
Hi @dstsimokha 

Thank you so much for your contribution. Could you please resolve the conflicts in your branch?

simpl1g:
@BentsiLeviav there is `persist_docs` macro that already adds comment to table and columns, I guess this PR should be reverted

PR #244, opened 2024-01-25
fix: use correct schema for MV target tables

SoryRawyer:
## Summary
The existing implementation of materialized views had one case where the target relation name was passed as a string of just the table name, instead of the full relation object. This meant that any custom schema would not be included in the final DDL statement, and the MV would break as it would try to write to a target table that didn't exist. This change passes the full relation object when getting the "create materialized view" SQL, which in turn provides the full path to the table in every case.

## Checklist

### DIFF
--- a/dbt/include/clickhouse/macros/materializations/materialized_view.sql
+++ b/dbt/include/clickhouse/macros/materializations/materialized_view.sql
@@ -41,7 +41,7 @@
       {{ clickhouse__get_create_materialized_view_as_sql(target_relation, sql) }}
     {%- endcall %}
   {% elif existing_relation.can_exchange %}
-    {{ log('Replacing existing materialized view' + target_relation.name) }}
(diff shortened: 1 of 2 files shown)


SoryRawyer:
Hey @BentsiLeviav, let me know if there's anything I can help out with on this PR! I looked at the failing test and it doesn't seem to be related to my changes. Everything ran successfully locally on my end.

BentsiLeviav:
Hi @SoryRawyer 
I'll give it a look. I want to run the tests locally on my end to make sure everything is ok. will update here.

BentsiLeviav (approved):
I ran the tests locally and it seems to be ok

bmtcril:
@BentsiLeviav we've been waiting on this fix, is there anything left to do here? I don't think @SoryRawyer can merge this.

Issue #223, opened 2023-12-17
Generic test - "relationships" issue after with test setting limit


zwbill82:
### Describe the bug
14 of 15 ERROR relationships_dm_est_unit_economics_dbt_listing_id__listing_id__ref_stg_dm_unit_product_pool_online_  [ERROR in 0.45s]
 Database Error in test relationships_dm_est_unit_economics_dbt_listing_id__listing_id__ref_stg_dm_unit_product_pool_online_ (models/dm_est_unit_economics/marts/_dm_est_unit_economics.yml)
  Code: 62.
  DB::Exception: Syntax error: failed at position 769 ('limit') (line 34, col 7): limit 100
      ) dbt_internal_test. Expected one of: UNION, EXCEPT, INTERSECT. Stack trace:
  
  0. DB::Exception::Exception(DB::Exception::MessageMasked&&, int, bool) @ 0xe11e335 in /usr/bin/clickhouse
  1. ? @ 0x8d1cc2d in /usr/bin/clickhouse
 
### Steps to reproduce
1. set global tests setting in dbt_project.yml
`tests:
  +warn_if: ">20"
  +error_if: ">50"
  +limit: 100`
2. add test like this:
` - name: listing_id
      tests:
        - unique:
            config:
              where: listing_id>0

        - relationships:
            to: ref('stg_dm_unit_product_pool_online')
            field: listing_id`

### Expected behaviour
<img width="1386" alt="image" src="https://github.com/ClickHouse/dbt-clickhouse/assets/29267289/0b8c0db9-7c99-45dc-baaa-34a8de07da95">

### Code examples, such as models or profile settings
the run code is here, the limit clause is after the setting clause.
`select
      count(*) as failures,
      count(*) >20 as should_warn,
      count(*) >50 as should_error
    from (
      
    
    

with child as (
    select listing_id as from_field
    from `dbt_dev`.`dm_est_unit_economics_dbt`
    where listing_id is not null
),

parent as (
    select listing_id as to_field
    from `dbt_dev`.`stg_dm_unit_product_pool_online`
)

select
    from_field

from child
left join parent
    on child.from_field = parent.to_field

where parent.to_field is null
settings join_use_nulls = 1

      limit 100
    ) dbt_internal_test`

#### Environment
* dbt version: 1.7.4
* dbt-clickhouse version: 1.7.1
* clickhouse-driver version (if using native): 0.2.6
* clickhouse-connect version (if using http):
* Python version: 3.8
[...]

PR #245, opened 2024-01-28
Bug/223 relationship test with limit

BentsiLeviav:
## Summary
As mentioned in issue #223, dbt fails to execute relationship tests when a limit clause is specified. The issue arises because the relationship test macro appends the join_use_nulls settings, and the limit is added at the end of the query (after the settings), causing a parsing error.

To address this limitation of adding a limit clause when the query includes settings, we developed a new macro that parses the query and inserts the limit just before the settings clause.

This is accomplished by introducing a flag within a comment, which is then used to parse the settings section accordingly.

## Checklist

### DIFF
--- a/dbt/adapters/clickhouse/impl.py
+++ b/dbt/adapters/clickhouse/impl.py
@@ -412,7 +412,14 @@ class ClickHouseAdapter(SQLAdapter):
         res = []
         for key in settings:
             res.append(f' {key}={settings[key]}')
-        return '' if len(res) == 0 else 'SETTINGS ' + ', '.join(res) + '\n'
+        if len(res) == 0:
+            return ''
+        else:
+            settings_str = 'SETTINGS ' + ', '.join(res) + '\n'
+            return f"""
+                        -- settings_section
+                        {settings_str}
+                        """
 
     @available
     def get_model_query_settings(self, model):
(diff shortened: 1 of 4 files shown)


genzgd:
Maybe we could always add an `-- end_of_sql` marker (regardless of whether there are settings), instead of a `-- settings_section` marker only if there are settings.  That would allow finding the last "-- end_of_sql" marker and insert the limit there, regardless of any nesting that may happen in building the query.

Since there aren't any real automated dbt tests around the `dbt test` command, has this been validated manually?

BentsiLeviav:
@genzgd Yes, `-- end_of_sql` is better. I changed the code.
I ran the tests, and everything seems to be OK besides 8 tests (which were broken even before my changes).

dbt/adapters/clickhouse/impl.py
```diff
@@ -412,15 +412,30 @@ def get_model_settings(self, model):
         res = []
         for key in settings:
             res.append(f' {key}={settings[key]}')
-        return '' if len(res) == 0 else 'SETTINGS ' + ', '.join(res) + '\n'
```
genzgd, line 415: Let's always add the `-- end_of_sql` marker comment regardless of whether we have actual settings to add?  That way we can check for it in other circumstances where we need to do stuff after the "main" sql.

genzgd (changes requested):

dbt/adapters/clickhouse/impl.py
BentsiLeviav: Do you mean to add that at the end of every macro?

genzgd: It would be the same macros where we add `settings` now, just _always_ add `-- end_of_sql` instead of conditionally.

I think the `settings` thing is fairly unique to ClickHouse so I feel like being able to consistently determine "this is the end of regular SQL" from "this might be ClickHouse doing nonstandard stuff" will lead to more straightforward code, even if we don't actually add the settings.  Note that there are so many possible query and table settings that we'll probably end up with them most of the time anyway.

BentsiLeviav: Done

BentsiLeviav: Opened a new PR in #249

Issue #246, opened 2024-02-04
Fix tests when running with ClickHouse cloud instance


BentsiLeviav:
### Describe the bug

When running the tests with a ClickHouse cloud instance, 3 tests are failing: `TestModelGrants.test_view_table_grants`, `TestSeedGrants.test_seed_grants` and `TestPersistDocsColumnMissing.test_missing_column`.

### Steps to reproduce
1. Configure your test.dev file with a ClickHouse cloud instance
2. Run `pytest tests`

### Expected behavior

All tests should pass.

BentsiLeviav:
All tests are passing now for the job `cloud_tests`.

Issue #231, opened 2023-12-28
settings is not passed to CREATE VIEW AS


lksv:
### Steps to reproduce

I need to pass settings to `CREATE VIEW AS`, specifically `CREATE VIEW ... AS (...) SETTINGS flatten_nested=0` .

However, when I create a model (view) as follows, the setting is not used.

```
{{
    config(
        materialized='view',
        settings={'flatten_nested': 0},
    )
}}
```

I suspect the solution is to add `{{ adapter.get_model_settings(model) }}` to the [clickhouse__create_view_as macro](https://github.com/ClickHouse/dbt-clickhouse/blob/ca9da0b6ebacc14d67139b0b417e0d4083b68a9c/dbt/include/clickhouse/macros/materializations/distributed_table.sql#L117) and also to [materialed_view](https://github.com/ClickHouse/dbt-clickhouse/blob/v1.7.1/dbt/include/clickhouse/macros/materializations/materialized_view.sql)

Currently, I use workaround with `pre_hook="SET flatten_nested=0"`. Howerever, as the [README](https://github.com/ClickHouse/dbt-clickhouse#set-statement-warning) mentions, it is not recomended to use `SET`.

### Expected behavior

* dbt version: 1.7.4
* dbt-clickhouse version: 1.7.1
* clickhouse-driver version (if using native): 0.2.6
* Python version: 3.10.12
* Operating system: Linux (ubuntu 23.10.)

#### ClickHouse server
* ClickHouse Server version: 23.8.8

BentsiLeviav:
Hi @lksv 
Just created a PR for solving this in #249

PR #251, opened 2024-02-18
Add snapshot materialization to default settings

BentsiLeviav:
Thanks to the test `test_snapshot_check_cols.TestSnapshotCheckCols` I found out I didn't add all materializations when changing default settings in #250

### DIFF
--- a/dbt/adapters/clickhouse/dbclient.py
+++ b/dbt/adapters/clickhouse/dbclient.py
@@ -103,6 +103,8 @@ class ChClientWrapper(ABC):
             "ephemeral": {},
             "materialized_view": {},
             "snapshot": {},
+            "distributed_table": {},
+            "distributed_incremental": {},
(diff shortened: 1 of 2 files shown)


genzgd:
Tests still aren't passing?

BentsiLeviav:
@genzgd While trying to solve #246, I found out `test_snapshot_check_cols.TestSnapshotCheckCols` fails as well (due to the changes I made in #231.

genzgd:
Understood, I meant that even with this fix the test suite still isn't passing.  That is on me, I shouldn't have approved #231 but I missed that tests were also failing there.

One problem is, like snapshot, it looks like `'distributed_table'` also isn't covered by the new settings code.  Even when I tried adding that locally I got test failures, so we need to dig a little deeper before releasing again.

genzgd (approved):

PR #267, opened 2024-04-08
feat: allow to add connection overrides for dictionaries

Ian2012:
## Summary
This PR fixed an issue in which ClickHouse was using the configured host to load the dictionary, however, this would fail if it was running on Docker and would be slower if it was using a URL to connect to ClickHouse. It implements a new config for the dictionary model allowing people to override connection params.

### Before
`SOURCE(CLICKHOUSE(USER 'ch_admin' PASSWORD '[HIDDEN]' HOST 'clickhouse' PORT '8123' QUERY ...)`

### After
`SOURCE(CLICKHOUSE(USER 'ch_admin' PASSWORD '[HIDDEN]' QUERY ...)`

## Checklist

### DIFF
--- a/dbt/adapters/clickhouse/impl.py
+++ b/dbt/adapters/clickhouse/impl.py
@@ -458,17 +458,24 @@ class ClickHouseAdapter(SQLAdapter):
         return [{'name': column.name, 'data_type': column.dtype} for column in columns]
 
     @available
-    def get_credentials(self) -> Dict:
+    def get_credentials(self, connection_overrides) -> Dict:
         conn = self.connections.get_if_exists()
         if conn is None or conn.credentials is None:
             return dict()
-        return {
+        credentials = {
             'user': conn.credentials.user,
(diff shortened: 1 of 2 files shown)


Ian2012:
@BentsiLeviav This PR is ready for review

rjoelnorgren:
@BentsiLeviav, I was literally about to create an issue asking for very similar change. Only difference is I think it might be nice to support omitting credentials altogether for the table source? i.e.

`SOURCE(CLICKHOUSE(TABLE 'table_name'))`

Specifying the user in our case fails in our development environment because developers are not able to directly authenticate to ClickHouse from the server itself. As a result we get an auth error.

SoryRawyer:
I think @rjoelnorgren brings up a good point. Would it be possible to update the `credentials` dict with `connection_overrides` and, if a given field is empty, omit that from the configuration?

genzgd (approved):

PR #285, opened 2024-05-29
Added database prefix option for local tables of distributed tables

### LINKED ISSUE #248: Question: Is it possible to set a separate database for local tables using `distributed_table` materialisation?

steffen030:
## Summary

Distributed Tables currently create their local tables within the same database. However, this is not always desirable as for other users of these tables it might not be clear which table to use. At the same time it is very different to apply seperate permissions to local tables. 

This PR creates the option to add a database prefix for local tables which allows for filtering and seperate permissions.

Closes #248 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ your_profile_name:
       use_lw_deletes: [False] Use the strategy `delete+insert` as the default incremental strategy.
       check_exchange: [True] # Validate that clickhouse support the atomic EXCHANGE TABLES command.  (Not needed for most ClickHouse versions)
       local_suffix [_local] # Table suffix of local tables on shards for distributed materializations.
+      local_db_prefix [local_] # Database prefix of local tables on shards for distributed materializations. If empty, it uses the same database as the distributed table.
       allow_automatic_deduplication [False] # Enable ClickHouse automatic deduplication for Replicated tables
       custom_settings: [{}] # A dictionary/mapping of custom ClickHouse settings for the connection - default is empty.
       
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -78,6 +78,7 @@ your_profile_name:
       use_lw_deletes: [False] Use the strategy `delete+insert` as the default incremental strategy.
       check_exchange: [True] # Validate that clickhouse support the atomic EXCHANGE TABLES command.  (Not needed for most ClickHouse versions)
       local_suffix [_local] # Table suffix of local tables on shards for distributed materializations.
+      local_db_prefix [local_] # Database prefix of local tables on shards for distributed materializations. If empty, it uses the same database as the distributed table.
```
maowerner, line 81: I believe the default value should be [] here

steffen030: Well spotted, thanks!

BentsiLeviav:
You also need to adjust `_ensure_database` at https://github.com/ClickHouse/dbt-clickhouse/blob/9c212f285a0ba8c30a3a0e32f389412400049dc7/dbt/adapters/clickhouse/dbclient.py#L199

steffen030:
> You also need to adjust `_ensure_database` at
> 
> https://github.com/ClickHouse/dbt-clickhouse/blob/9c212f285a0ba8c30a3a0e32f389412400049dc7/dbt/adapters/clickhouse/dbclient.py#L199

@BentsiLeviav It seems like I missed ensuring that the database existed, but `_ensure_database` doesn't seem to work nicely for that as it checks it only at adapter level. However, the schema/ database can be changed on model level and it wouldn't work there. I added some adjustments to ensure the database gets created.

Issue #295, opened 2024-06-13
Table and Column Comments set up misbehaviour


GavrG:
### Describe the bug
Testing [v1.7.7 Community PRs](https://github.com/ClickHouse/dbt-clickhouse/releases/tag/v1.7.7) 
exacly: allows to add a comment in table's or view's metadata by @dstsimokha in https://github.com/ClickHouse/dbt-clickhouse/pull/228
We've found, that generated ALTER TABLE SQL "ignores" ON CLUSTER clause and all changes done only on connection node.
Clickhouse and DBT does not return any errors.

### Steps to reproduce
1. Create a model: models\test_comment_model.sql
```
{{
  config
      (
        materialized = 'table',
        schema = 'dbt__core',
        order_by=('some_id'),
        engine='ReplicatedMergeTree',
        persist_docs={"relation": true, "columns": true}
      )
}}

Select
    1 AS some_id,
    '----' AS some_column_name,
    '++++' AS some_other_column_name
 ```
2. Create a description: models\models.yml
```
version: 2
models:
  - name: test_comment_model
    description: | 
      test_comment_model description
    columns:
      - name: some_id
        description: some_id description
      - name: some_column_name
        description: some_column_name description
      - name: some_other_column_name
        description: some_other_column_name description
```
3. Build model dbt build  --select test_comment_model.sql
4. Checking SQL in dbt_logs
```
  alter table `dbt__core`.`test_comment_model` 
    ON CLUSTER "clust"  
  modify comment $dbt_comment_literal_block$test_comment_model description$dbt_comment_literal_block$, 
  comment column some_id  $dbt_comment_literal_block$some_id description$dbt_comment_literal_block$, 
  comment column some_column_name  $dbt_comment_literal_block$some_column_name description$dbt_comment_literal_block$, 
  comment column some_other_column_name $dbt_comment_literal_block$some_other_column_name description$dbt_comment_literal_block$
```
5. Check it up in Clickhouse
5.1. columns comments
```
SELECT hostName() as hostName,`database`, name, `comment`, *
FROM clusterAllReplicas('{cluster}', system.columns)
where database = 'dbt__core'
and `table` = 'test_comment_model'
order by database,  name, hostName
```
hostName      |database |name                                   |comment
clickhouse-01|dbt__core|some_column_name            |some_column_name description 
clickhouse-02|dbt__core|some_column_name            | 
clickhouse-01|dbt__core|some_id                               |some_id description  
[...]

PR #313, opened 2024-07-05
Update persist_docs.sql

GavrG:
Fixes Issue Table and Column Comments set up misbehaviour [#295](https://github.com/ClickHouse/dbt-clickhouse/issues/295)


BentsiLeviav:
Hi @GavrG 

Thank you for your contribution.
Why splitting the commends is supposed to solve the issue?
The suggested solution doesn't resolve the problem as it is a ClickHouse issue and not a dbt (adapter) issue (the responsibility of syncing nodes' metadata).

The workaround I would suggest is to use `clusterAllReplicas` in the macro, which queries the entire cluster.

PR #323, opened 2024-07-18
Adding new unique key column - fixes #322

### LINKED ISSUE #322: Cannot add a new column to incremental model that is also a unique_key
If I create an incremental model, and then add a column which is also added to the `unique_key`, the model fails to build with the error : 

```
DB::Exception: Missing columns: 'col_3' while processing query: 'SELECT col_1, col_2 FROM dbt_clickhouse_2760_test_schema_change_1721300786653.schema_change_append WHERE (col_1, col_3) NOT IN ((SELECT col_1, col_3 FROM dbt_clickhouse_2760_test_schema_change_1721300786653.schema_change_append__dbt_new_data) AS _subquery18)', required columns: 'col_1' 'col_2' 'col_3', maybe you meant: 'col_1' or 'col_2'. (UNKNOWN_IDENTIFIER) (version 23.6.2.18 (official build))
```

It seems to be failing when trying to insert the records from the old table that are not in the `new_data` table.

See related PR for unit test that demonstrates the issue:  https://github.com/the4thamigo-uk/dbt-clickhouse/pull/1/files

I wonder though, how this could be fixed?

One way would be to modify the query as indicated : 

```
SELECT   col_1,   col_2 
FROM 
  dbt_clickhouse_2760_test_schema_change_1721300786653.schema_change_append 
WHERE 
    (
      col_1,
      col_3  <--- REMOVE THE NEW UNIQUE KEY COLUMN
    ) NOT IN (
    (
      SELECT
        col_1,
        col_3   <--- REMOVE THE NEW UNIQUE KEY COLUMN
      FROM 
        dbt_clickhouse_2760_test_schema_change_1721300786653.schema_change_append__dbt_new_data
    ) AS _subquery18
  )
```

Do you think this approach is viable/correct?

Additionally: I notice that for schema changes the strategy switches to the legacy approach, is there scope to implement this using `delete+insert`?

the4thamigo-uk:
## Summary
Proposed solution to support adding new columns that are also added to the unique_key.

### DIFF
--- a/dbt/include/clickhouse/macros/materializations/incremental/incremental.sql
+++ b/dbt/include/clickhouse/macros/materializations/incremental/incremental.sql
@@ -154,12 +154,23 @@
     -- table.
     {%- set source_columns = adapter.get_columns_in_relation(existing_relation) -%}
     {%- set source_columns_csv = source_columns | map(attribute='quoted') | join(', ') -%}
+
+    -- Existing table does not have any of the new columns that have been added to the unique_key, so we remove them
+    {% set old_unique_key = [] %}
+    {%- for source_column in source_columns -%}
+        {%- if source_column.name in ( unique_key.split(',') | map('trim') ) -%}
+            {{ old_unique_key.append(source_column.quoted) }}
+        {%- endif -%}
+    {%- endfor -%}
+
+    {%- set old_unique_key_csv = old_unique_key | join(', ') -%}
+
     {% call statement('insert_existing_data') %}
         insert into {{ inserted_relation }} ({{ source_columns_csv }})
         select {{ source_columns_csv }}
         from {{ existing_relation }}
-          where ({{ unique_key }}) not in (
-            select {{ unique_key }}
+          where ({{ old_unique_key_csv }}) not in (
+            select {{ old_unique_key_csv }}
             from {{ inserting_relation }}
           )
        {{ adapter.get_model_query_settings(model) }}
--- a/tests/integration/adapter/incremental/test_schema_change.py
+++ b/tests/integration/adapter/incremental/test_schema_change.py
@@ -5,7 +5,7 @@ schema_change_sql = """
 {{
     config(
         materialized='incremental',
-        unique_key='col_1',
+        unique_key='col_1' + var('extra_unique_keys',''),
         on_schema_change='%schema_change%'
     )
 }}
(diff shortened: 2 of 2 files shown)


the4thamigo-uk:
fixes #322

BentsiLeviav:
Hi @the4thamigo-uk 

This https://github.com/ClickHouse/dbt-clickhouse/pull/332 was opened a few days ago and handles schema changes. Generally, support in the `sync_all_columns` option was introduced.
Could you check if it solves your issue?

Thanks

the4thamigo-uk:
@BentsiLeviav Ah yes I noticed the legacy fallback, but I was aiming to do a minimal fix. Great that there is a better solution.

I suppose if the PR passes a test which is equivalent to https://github.com/ClickHouse/dbt-clickhouse/pull/323/files#diff-f2e636456ac1c2366b5fd10e6beaf3560a946882645f7cee6ba0e0501bf301e3 then it would be fine.

the4thamigo-uk:
> I see that the table creation passes successfully, but is there a chance there is a mistake in [this](https://github.com/ClickHouse/dbt-clickhouse/pull/323/files#diff-f2e636456ac1c2366b5fd10e6beaf3560a946882645f7cee6ba0e0501bf301e3R90) assertion? After running the test the table looks like this: image

Yes its interesting to know what the behaviour should be I suppose. I guess you are suggesting row 4 shuld not be present?

BentsiLeviav:
@the4thamigo-uk 
What ClickHouse version do you use?

the4thamigo-uk:
> @the4thamigo-uk What ClickHouse version do you use?

24.3.4.147 - but would be surprised if it makes a difference...

BentsiLeviav:
That is really weird. Did you test it with the changes contributed at #332 ?
At the beginning, the table has the following data:
image

and the incremental query:
```SQL

select
    number as col_1,
    number + 1 as col_2,
    number + 2 as col_3
from numbers(2, 3)
```

 yields this:

image

So the append of these 2 groups would be 6 rows:
image


Anyway, the error mentioned in #322 doesn't appear anymore. I'm closing this PR and the issue, feel free to comment/reopen if needed or if something is wrong.

the4thamigo-uk:
Yeah weird... anyway, the incremental schema changes is much better. Thanks @canbekley

PR #324, opened 2024-07-20
feat: support query_settings in view materialization

james-seymour-cubiko:
## Summary

Currently the `settings` configuration option is used when creating a view, but only at view **creation** time. These settings do not persist when actually querying the view, so this PR adds the option to use the separate `query_settings` config option to create a view with settings that persist at query time.

My specific use case is to allow `join_use_nulls=1` to be set within dbt, rather than remembering to include this setting every time I query the view.   

I'm a first timer here, so have a few questions too,
- There's no open issue yet - should I create one?
- I didn't see any tests that were relevant to this change, but if there are some that cover this just lemme know and I'll add some in :)
- Would also be great to know what specific docs would need to change if this is merged in?

### DIFF
--- a/dbt/include/clickhouse/macros/materializations/view.sql
+++ b/dbt/include/clickhouse/macros/materializations/view.sql
@@ -78,6 +78,7 @@
     {%- endif %}
   as (
     {{ sql }}
+    {{ adapter.get_model_query_settings(model) }}
   )
       {% if model.get('config').get('materialized') == 'view' %}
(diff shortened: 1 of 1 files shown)


BentsiLeviav:
Hi @james-seymour-cubiko 
Thanks for your contribution!

Per your question, I don't see any particular test that includes this change. 
Regarding the issue - it is ok, you don't have to create one.

I ran the tests and all passed, therefore I'm merging this change.

PR #342, opened 2024-08-18
Support projections

### LINKED ISSUE #319: Adding PROJECTION to models
**Is your feature request related to a problem? Please describe.**
It would be awesome if we could add [projections](https://clickhouse.com/docs/en/sql-reference/statements/alter/projection) to models, this is very helpful when using a semantic layer on top of the data marts, to facilitate and speed up querying for certain filters.

**Describe the solution you'd like**
IMHO, the ideal solution would be to configure the projection as a `config` parameter (not sure though if dbt supports this):

```sql
-- models/marts/touchpoints;
{{
    config(
        materialized='materialized_view',
        engine='MergeTree()',
        order_by='workspace_id, toStartOfHour(updated_at), touchpoint_id',
        projections=[
            { "name": "some_projection", "select": ["touchpoint_id", "some_other_col"], "order_by": ["touchpoint_id"] }
        ]
    )
}}
select * from {{ source('other', 'touchpoints') }}
```

**Describe alternatives you've considered**
The alternative that we use currently is using a migration script to put the projection on various models, after they have been built, which is not really nice, since the code creating a projection and the model are separate from each other and we need to take care that the database migrations are always up to date with the dbt models.

BentsiLeviav:
## Summary
close #319 

This PR introduces support for projections in `dbt-clickhouse`. Projections are added to the `table` and `distributed_table` materializations as a model setting. For distributed tables, the projection is applied to the `_local` tables, not to the distributed proxy table.

## Tests
I added a test for each materialization. During test development, I encountered two challenges:

1. To verify that the projection was used by the query, we need to query `system.query_log`. However, this table takes time to create, which necessitates adding a sleep statement.
2. Currently, there's no straightforward way to pass a `query_id` parameter to ClickHouse, making it difficult to retrieve the exact query I want to check. As a workaround, I queried `system.query_log` using a `LIKE` statement.

### DIFF
--- a/dbt/include/clickhouse/macros/materializations/table.sql
+++ b/dbt/include/clickhouse/macros/materializations/table.sql
@@ -147,8 +147,26 @@
         {% call statement('create_table_empty') %}
             {{ create_table }}
         {% endcall %}
+         {% if config.get('projections') is not none%}
+             {% call statement('add_projections') %}
+                {{ projection_statement(relation) }}
+            {%endcall  %}
+         {% endif %}
+
+
         {{ clickhouse__insert_into(relation, sql, has_contract) }}
     {%- endif %}
+{%- endmacro %}
+
+{% macro projection_statement(relation) %}
+    {%- set projections = config.get('projections', default=[]) -%}
+
+    {%- for projection in projections %}
+        ALTER TABLE {{ relation }} ADD PROJECTION {{ projection.get('name') }}
+        (
+            {{ projection.get('query') }}
+        )
+    {%- endfor %}
 {%- endmacro %}
 
 {% macro create_table_or_empty(temporary, relation, sql, has_contract) -%}
--- a/dbt/include/clickhouse/macros/materializations/table.sql
+++ b/dbt/include/clickhouse/macros/materializations/table.sql
@@ -147,7 +147,7 @@
         {% call statement('create_table_empty') %}
             {{ create_table }}
         {% endcall %}
-         {% if config.get('projections') is not none%}
+         {% if config.get('projections')%}
              {% call statement('add_projections') %}
                 {{ projection_statement(relation) }}
             {%endcall  %}
--- a/tests/integration/adapter/projections/test_projections.py
+++ b/tests/integration/adapter/projections/test_projections.py
@@ -1,5 +1,6 @@
 import os
 import time
+import uuid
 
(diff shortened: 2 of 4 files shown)


dbt/include/clickhouse/macros/materializations/table.sql
```diff
@@ -147,10 +147,28 @@
         {% call statement('create_table_empty') %}
             {{ create_table }}
         {% endcall %}
+         {% if config.get('projections') is not none%}
```
genzgd, line 150: In the previous file you don't have the "not none" piece.  I think it's better without, assuming the `if` evaluates as false for both `None` and empty arrays.

genzgd (approved):

tests/integration/adapter/projections/test_projections.py
```diff
@@ -0,0 +1,116 @@
...
+        result = project.run_sql(query, fetch="all")
+        assert len(result) == 3  # We expect 3 departments in the result
+        assert result == [('engineering', 43.666666666666664), ('malware', 40.0), ('sales', 25.0)]
+
+        # waiting for system.log table to be created
+        time.sleep(10)
+
+        # check that the latest query used the projection
+        result = project.run_sql(
+            f"SELECT query, projections FROM system.query_log WHERE query like '%{query}%' ORDER BY query_start_time DESC",
```
mshustov, line 80: clusterAllReplicas(default, 'system.query_log')

BentsiLeviav:
Writing it here for community awareness - the feature was added both for `table` and `distributed_table` materialization, but it doesn't fully function with `distributed_table`. I marked the distributed test with `@pytest.mark.xfail`.

arun11299:
@BentsiLeviav Would this be a problem when using Clickhouse Cloud ?

BentsiLeviav:
@arun11299 it is supposed to work on ClickHouse Cloud with no problems.
For an example please refer to this changelog update
https://github.com/ClickHouse/dbt-clickhouse/blob/main/CHANGELOG.md#release-182-2024-08-22

MarekVigas:
@BentsiLeviav I think it would be nice to at least mention it in README.md under [supported-features](https://github.com/ClickHouse/dbt-clickhouse?tab=readme-ov-file#supported-features) I really had to dig if projections are supported or not.

PR #347, opened 2024-08-26
Create CONTRIBUTING.md

BentsiLeviav:


CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Getting Started
+
+### 1. Fork the Repository
+
+Start by forking the repository on GitHub. This will create a copy of the repository under your own GitHub account.
+
+### 2. Set Up Environment
+
+* Make sure Python is installed locally.
```
mshustov, line 16: let's add a pointer to specific python version required

CONTRIBUTING.md
```diff
@@ -0,0 +1,76 @@
...
+your changes, and adjust it. If not such test, please create one.
+
+### 6. Run The Tests
+
+See [Running Tests](#running-tests) for more information.
+
+> **Important:** Please make sure the tests are running successfully before pushing your code.
+
+### 7. Create a PR
+Create a pull request from your forked repository to the main one. In case this is your first contribution, make sure to sign ClickHouse's CLA.
```
mshustov, line 57: optional: ask to add a sensible description of a feature and the use case

mshustov (approved):

CONTRIBUTING.md
simpl1g: I guess It would be good idea to stick to some commit message format, like in https://www.conventionalcommits.org for example. Like `fix: ...`, `docs: ...`

CONTRIBUTING.md
BentsiLeviav: Done

CONTRIBUTING.md
BentsiLeviav: Done

PR #362, opened 2024-09-25
Apply query settings for empty table model

alexsubota:
## Summary
Fix for query_settings
Currently, when a table is created, the query settings are not passed to the query, which causes an issue with the join_use_nulls setting (empty table have not nullable column, and then we cant insert null into this column). This PR fixes this behavior.

### DIFF
--- a/dbt/include/clickhouse/macros/materializations/table.sql
+++ b/dbt/include/clickhouse/macros/materializations/table.sql
@@ -203,6 +203,7 @@
             {{ sql }}
           )
         {%- endif %}
+        {{ adapter.get_model_query_settings(model) }}
     {%- endif %}
 
 {%- endmacro %}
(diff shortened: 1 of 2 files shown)


genzgd:
Is it possible to add a test case?  It's probably okay but can't tell by looking whether (for example), the contract logic interferes with the query.

alexsubota:
i can create test model like this
```
{{
    config(
        materialized='table',
        query_settings={
            'join_use_nulls': 1
        }
    )
}}
select t2.id as test_id
from (select 1 as id) t1
         left join (select 2 as id) t2
on t1.id=t2.id
```
and check that test_id has a nullable type.
Is it ok?

alexandr-subota-idf:
@genzgd check test please

alexsubota:
@genzgd check test please

BentsiLeviav:
@genzgd I checked locally and it doesn't collide with the contract (from a syntax point of view).

BentsiLeviav (approved):

PR #363, opened 2024-10-01
don't drop local tables when there is no distributed table

canbekley:
## Summary
Current distributed incremental runs will drop local production tables when the distributed tables are missing. This can cause irrevesable data loss.

The proposed fix will only drop and create local tables if they actually don't exist in clickhouse. Distributed tables will be created if they are missing (for any reason).

## Checklist


canbekley:
@BentsiLeviav I added a changelog for this as well

Issue #280, opened 2024-05-01
Two materialized views writing to the same table


the4thamigo-uk:
CH MVs are insert triggers on the source table. It is possible to setup two or more MVs that write their records to a single table. from what I can tell the dbt MV implementation creates a destination table for each model.

**Describe the solution you'd like**
Is it possible to suppress somehow map a model to a different table so that it can be shared between two or more MV models?

the4thamigo-uk:
I was thinking about this, although it sounds fairly easy to implement, its kinda a bit messy when it comes to deciding when to backfill with data from the second `materialized_view` model. Currently it backfills when the destination table does not exist. However, by the time the second model is built the destination model has already been built, so you cant use this method to decide to backfill. It seems that the second model needs to know if the first model performed a backfill in order to decide whether the second model should do a backfill.

An alternative approach might be to extend the `materialized_view` materialization code to somehow manage more than one materialized view. However, I think dbt really only supports specifying a single SQL statement so it is kinda difficult to see how this would work...

the4thamigo-uk:
Here is an idea : 

```
{{
  config(
     materialized='materialized_view',
     engine='ReplicatedSummingMergeTree',
     order_by='id'
     views=['mv1','mv2'],
  ) 
}}

--begin:mv1
select
    1 as id,
    sum(value) as total_value
from
    {{ ref("source1") }}
group by all
--end:mv1

union all

--begin:mv2
select
    1 as id,
    sum(value) as total_value
from
    {{ ref("source2") }}
group by all
--end:mv2
```

Basically, we have a single model that encapsulates SQL for any number of clickhouse materialized views that target the same destination table. Historical data is processed by retrieving the rows of the union of all sql statement blocks, but when each of the the materialized views are built, the materialization should extract the only the sql between the begin and end tags. The names of the materialized views ('mv1' and 'mv2') are explicitly specified in a config option, and each of these should have a single sql begin/end block.

The advantage of this is that we dont have the issues above regarding how to populate the historical data as we are just using a single query to do this at table creation time (just like how it works at the moment). Secondly, it nicely encapsulates all the sql relating to the final destination table so it is easier to understand what is populating it. Lastly, we retain all the linkage from and `ref()` and `source()` functions used in the body sql.

PR #364, opened 2024-10-08
Allow multiple materialized views to write to same target (#280)

the4thamigo-uk:
## Summary
Proposed mechanism to support having multiple materialized views pointing to the same target table.

In brief, we add tags to mark blocks of the sql which will form each of the materialized views to be created. 

i.e. https://github.com/ClickHouse/dbt-clickhouse/pull/364/files#diff-b1cddf2e5c9b3d08ea2276e5cef4944b06b9744a4076c5970ff5a5edb0140f08R62.

The entire union of all such blocks is used to backfill the target table.

## Checklist

### DIFF
--- a/tests/integration/adapter/materialized_view/test_materialized_view.py
+++ b/tests/integration/adapter/materialized_view/test_materialized_view.py
@@ -15,6 +15,9 @@ id,name,age,department
 1231,Dade,33,engineering
 6666,Ksenia,48,engineering
 8888,Kate,50,engineering
+1000,Alfie,10,sales
+2000,Bill,20,sales
+3000,Charlie,30,sales
 """.lstrip()
 
 # This model is parameterized, in a way, by the "run_type" dbt project variable
@@ -40,8 +43,7 @@ select
 from {{ source('raw', 'people') }}
 where department = 'engineering'
 
-{% else %}
-
+{% elif var('run_type', '') == 'extended_schema' %}
 select
     id,
     name,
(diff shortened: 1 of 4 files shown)


the4thamigo-uk:
issue #280

BentsiLeviav:
Hi @the4thamigo-uk 

Before reviewing your PR, it is required to add a short description with a PR link to the changelog (please keep the current format we have in the Changelog file).
With this kind of feature, I think it is also important to add a note in the documentation (readme file).

looking forward to reviewing this

the4thamigo-uk:
@BentsiLeviav Ok, Ive done this.

tests/integration/adapter/materialized_view/test_materialized_view.py
```diff
@@ -191,3 +220,78 @@ def test_update_full_refresh(self, project):
...
+            "schema.yml": SEED_SCHEMA_YML,
+        }
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "hackers.sql": MV_MODEL,
+        }
+
+    def test_create(self, project):
```
BentsiLeviav, line 242: Could you also test for updates and full refresh?

BentsiLeviav:
Thanks @the4thamigo-uk
We appreciate this contribution!
The code LGTM. I left a small comment regarding the tests.

tests/integration/adapter/materialized_view/test_materialized_view.py
the4thamigo-uk: Yeah I think if you change the model to add or rename MVs, then old ones will be left running, which is incorrect.

BentsiLeviav:
I’ve added the tests through your branch.

Regarding the potential issue of unremoved materialized views (MVs): as you mentioned, this situation occurs when a user renames one (or more) of the MVs. What I’ve implemented for now is a query in ClickHouse to identify any MV that matches the pattern of the model name but is not listed in the model. If such an MV is found, a warning is raised to notify the user.

I’m hesitant about automatically dropping these MVs, as they could be valid and intentionally separate from the model.

Additionally, I’ve updated the README file to include this important information.

Let’s monitor this behavior for now. If anyone in the community has a better or safer solution, contributions are welcome!

the4thamigo-uk:
Amazing, thanks for finishing this off. I was actually working on this yesterday and implemented a similar thing to drop the dangling views, but as you say this might be dangerous.

However, I did change the naming convention for the views so that the view name is always appended after a `_mv_`. I think in this way it is a bit safer to drop anything that is prefixed with `hackers_mv_`

`hackers_mv1` -> `hackers_mv_mv1`

PR #353, opened 2024-09-05
add uuid's to table and view names

canbekley:
## Summary
Currently, a dbt run could drop temporary distributed tables or views while another run is still using them. This can lead to pipeline failure or even mutation failure (with delete+insert strategy). We want to avoid these issues by adding uuid's to table names.

## Checklist
Delete items not relevant to your PR:
- [ ] Unit and integration tests covering the common scenarios were added
- [ ] A human-readable description of the changes was provided to include in CHANGELOG
- [ ] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials

Issue #150, opened 2023-05-11
Randomize temp table __dbt_new_data name to allow concurrent dbt-clickhouse jobs


winterrobert:
**Is your feature request related to a problem? Please describe.**
We are running into an issue with the dbt-clickhouse adapter when using `delete+insert`. When running multiple dbt-clickhouse jobs at the same time, we often get a 500 complaining that the temporary xxx_dbt_new_data already exists:
```
09:36:29    :HTTPDriver for clickhouse:8123 returned response code 500)
09:36:29     Code: 57. DB::Exception: Table default.tablenamet__dbt_new_data already exists. (TABLE_ALREADY_EXISTS) (version 23.3.2.37 (official build))
```
My guess is that it's related to [this row here](https://github.com/ClickHouse/dbt-clickhouse/blob/main/dbt/include/clickhouse/macros/materializations/incremental.sql#L181)
```
{% macro clickhouse__incremental_delete_insert(existing_relation, unique_key, incremental_predicates) %}
    {% set new_data_relation = existing_relation.incorporate(path={"identifier": model['name'] + '__dbt_new_data'}) %}
    {{ drop_relation_if_exists(new_data_relation) }}
....
{% endmacro %}
```

**Describe the solution you'd like**
Would it be a good PR to add a random suffix to the table so that the temp tables created by dbt-clickhouse doesn't collide with each other? Could somebody at the team fix this (seems like a quick fix) or would you accept a PR?

**Describe alternatives you've considered**
Change our pipelines to make sure we always just run one dbt-clickhouse job at a time.

genzgd:
Included in the 1.4.1 patch release

PR #365, opened 2024-10-09
Dont drop tables in incremental append runs

stephen-up:
Hi, when dbt runs for a incremental+append it always drops the temporary tables that are used for full refreshes. Even if its not going to use them.

In my system, I have scheduled incremental runs of dbt that run frequently updating incremental+append tables. Whenever I do a full refresh (concurrently), the scheduled run drops the tables of the full refresh causing it to crash.

This PR stops that, by making the incremental+append runs not drop tables.

I see there have been a few other issues similar to this, but this seems like an easy minimal way to fix the issue.

- https://github.com/ClickHouse/dbt-clickhouse/issues/150
- https://github.com/ClickHouse/dbt-clickhouse/pull/353

Thanks

### DIFF
--- a/dbt/include/clickhouse/macros/materializations/incremental/incremental.sql
+++ b/dbt/include/clickhouse/macros/materializations/incremental/incremental.sql
@@ -22,7 +22,7 @@
   {%- set preexisting_backup_relation = load_cached_relation(backup_relation) -%}
   {% set incremental_strategy = adapter.calculate_incremental_strategy(config.get('incremental_strategy'))  %}
   
-  {% if not full_refresh_mode and incremental_strategy == 'append' %}
+  {% if not full_refresh_mode and incremental_strategy == 'append' and config.get('incremental_append_skip_drop_table', default='false') == 'true' %}
     -- skip dropping tables for incremental append runs.
   {% else %}
     {{ drop_relation_if_exists(preexisting_intermediate_relation) }}
--- a/dbt/include/clickhouse/macros/materializations/incremental/incremental.sql
+++ b/dbt/include/clickhouse/macros/materializations/incremental/incremental.sql
@@ -20,9 +20,14 @@
   {%- set backup_relation = make_backup_relation(target_relation, backup_relation_type) -%}
   {%- set preexisting_intermediate_relation = load_cached_relation(intermediate_relation)-%}
   {%- set preexisting_backup_relation = load_cached_relation(backup_relation) -%}
-
-  {{ drop_relation_if_exists(preexisting_intermediate_relation) }}
-  {{ drop_relation_if_exists(preexisting_backup_relation) }}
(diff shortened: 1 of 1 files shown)


stephen-up:
I signed the CLA btw. Dont know how long it takes to update.

BentsiLeviav:
Hi @stephen-up,

Thank you for your contribution! I appreciate the effort you’ve put into this.

I'm hesitant to approve this, as it could leave a hanging table behind. While I understand that it would be dropped in the next run, it’s still something to consider.

Additionally, this solution seems to address a specific edge case rather than providing a more comprehensive approach. A broader solution for handling parallel runs would be preferable.

I'll review the suggested alternative, though it currently applies only to temporary tables and not to relations created via make_intermediate_relation. Perhaps we could override make_intermediate_relation to include the invocation_id, similar to what was done in #150 and #353?

We could also introduce a feature flag for this macro, defaulting to false (disabled). What do you think?

stephen-up:
Hi Bentsi.

Leaving hanging tables does sound like something we want to avoid. Having random ids in the names like #150 or #353 do sound like solutions that ensure better concurrency, support multiple concurrent runs, but they increase the risk / cost of hanging tables.

Just not dropping the shadow tables that arent even used during an incremental+append seems simpler and more desirable for me. Im happy to add a feature flag for it.

stephen-up:
I've added the feature flag. Is there a way to set the feature flag for every model in a dbt project? They way i've implemented it, it looks like it needs to be set in every model like...

```sql
{{ config(
  order_by='project_id, igid',
  engine='ReplacingMergeTree(version)',
  materialized='incremental',
  incremental_strategy='append',
  incremental_append_skip_drop_table='true'
) }}
```

BentsiLeviav:
The feature flag I referred to in my previous reply was for the suggested `make_intermediate_relation` overwrite (my mistake if it wasn't clear). 
I created a new issue #420 to gather the discussion around this because multiple PRs and users are trying to solve this problem. Would you mind adding your thoughts there?

stephen-up:
Hi @BentsiLeviav, I've added some comments to #420. When thinking about race conditions, it gets tricky.

Do you want to support concurrent full refreshes. Or a pair of a refresh and an incremental run. If you just want the latter, like me. I think this PR is perfect.

The incremental + refresh seems like a common pattern. We've setup a pipeline where we have an incremental run every 1/2 hour then occasionally run full refreshes which take hours, easily leading to a clash.

koletzilla:
hi @stephen-up ! Sorry for coming back to your PR after so much time. We definetively need to work on the concurrency issue and I feel like #420 goes in the right direction. Just skipping here which tables to drop/not drop would create further consistency issues in the future as different combinations of dbt ops may happen that would reuse these not-dropped tables.

I'm going to close this PR in favour of the suggestions made in #420

Issue #366, opened 2024-10-10
Invalid setting when creating temporary tables with `ON CLUSTER`


shachibista:
Hi!

When trying to create a temporary table with the `cluster` property set, creating temporary tables with `Engine=Memory` passes the `replicated_deduplication_window` setting which is invalid. It seems that the `replicated_deduplication_window` setting only checks for [the materialization type](https://github.com/ClickHouse/dbt-clickhouse/blob/51bf0e1279b89856fdf2a50791d50e2b85b1fe18/dbt/adapters/clickhouse/dbclient.py#L115-L116) but does not check the `Engine` which it should.

I think that the [`update_model_settings`](https://github.com/ClickHouse/dbt-clickhouse/blob/51bf0e1279b89856fdf2a50791d50e2b85b1fe18/dbt/adapters/clickhouse/dbclient.py#L153-L159) method should accept engine type and only allow settings that are available for each engine.

Seeing this issue when trying to run https://github.com/elementary-data/dbt-data-reliability because it tries to create a [temporary table](https://github.com/elementary-data/dbt-data-reliability/blob/d3777b95230df68579705a6088e09d846b21f0db/macros/utils/table_operations/delete_and_insert.sql#L9) when artifacts change.

### Steps to reproduce
1. Setup a clickhouse multi-node cluster.
2. Configure with `cluster: "{cluster}"' in profiles
3. Run the elementary bootstrap model `dbt run -m elementary`

### Expected behaviour

No errors should be expected. The temporary table should be created in the connected node in memory.

### Code examples, such as models or profile settings

```yaml
# profiles.yml
project_name:
  target: local
  outputs:
    local:
      type: clickhouse
      driver: native
      schema: "{{ env_var('DBT_SCHEMA') }}"
      user: "{{ env_var('DBT_USER') }}"
      password: "{{ env_var('DBT_PASS') }}"
      port: 9000
      cluster: "{cluster}"
      cluster_mode: true
      check_exchange: false
```

```yaml
# packages.yml
packages:
  - package: dbt-labs/dbt_utils
    version: 1.2.0

packages:
  - package: elementary-data/elementary
    version: 0.15.2
```

```yaml
# dbt_project.yml
name: 'project_name'
version: '1.0.0'

profile: 'project_name'

model-paths: ["models"]
analysis-paths: ["analyses"]
test-paths: ["tests"]
seed-paths: ["seeds"]
macro-paths: ["macros"]
[...]

PR #367, opened 2024-10-11
Filter settings based on Engine

shachibista:
This PR introduces a feature that ignores settings that are not compatible with the Engine. More details in this issue: https://github.com/ClickHouse/dbt-clickhouse/issues/366

## Checklist

### DIFF
--- a/dbt/adapters/clickhouse/impl.py
+++ b/dbt/adapters/clickhouse/impl.py
@@ -480,10 +480,6 @@ class ClickHouseAdapter(SQLAdapter):
     def filter_settings_by_engine(self, settings, engine):
         filtered_settings = {}
 
-        if engine not in IGNORED_SETTINGS:
-            # If the engine has no settings it will not be in the IGNORED_SETTINGS map.
-            return filtered_settings
-
         if engine.endswith('MergeTree'):
             # Special case for MergeTree due to all its variations.
             ignored_settings = IGNORED_SETTINGS.get('MergeTree', [])
--- a/dbt/adapters/clickhouse/impl.py
+++ b/dbt/adapters/clickhouse/impl.py
@@ -131,24 +131,16 @@ ENGINE_SETTINGS = {
         "lightweight_mutation_projection_mode",
         "deduplicate_merge_projection_mode",
         "min_free_disk_bytes_to_perform_insert",
-        "min_free_disk_ratio_to_perform_insert"
+        "min_free_disk_ratio_to_perform_insert",
     ],
-    'Memory': [
-        'min_bytes_to_keep',
-        'max_bytes_to_keep',
-        'min_rows_to_keep',
-        'max_rows_to_keep'
-    ],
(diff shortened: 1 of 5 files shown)


BentsiLeviav:
@genzgd suggested changing it from a white list to a black list (remove settings we identified as not relevant for this engine), and in case we face a new engine, pass the settings as is.

@shachibista would you mind changing the logic to support this?

BentsiLeviav (changes requested):
Let me know WDYT about changing it to a black list instead of a white list

shachibista:
> Let me know WDYT about changing it to a black list instead of a white list

I don't personally agree with a black-list since:

1. It also requires some maintainance so the above reasoning about maintainence is moot.
2. It requires that invalid settings to be [discovered by users](https://github.com/ClickHouse/dbt-clickhouse/issues/366#issuecomment-2573740072) for each engine, forcing them to wait for the next release of the library.

However I have made the requested changes if it feels right to you since the original issue I was trying to solve was solved using another method (wrote own observability package, for posterity) and as such this issue is no longer important to me at this moment.

Hope that helps!

BentsiLeviav:
> It also requires some maintainance so the above reasoning about maintainence is moot.

There is a significant difference between updating the extensive settings dictionary with every ClickHouse (CH) release and allowing users to be completely blocked from using new configurations until we update it. By addressing mismatched settings, we can handle far fewer scenarios compared to CH introducing tens of new configurations every year.

> forcing them to wait for the next release of the library.

In my opinion, this approach is better than blocking the entire community from accessing new features or settings, which happens with every release.


> However I have made the requested changes if it feels right to you since the original issue I was trying to solve was solved using another method (wrote own observability package, for posterity) and as such this issue is no longer important to me at this moment.



Thank you for changing the code. Your efforts are sincerely appreciated and not taken for granted.

dbt/adapters/clickhouse/impl.py
```diff
@@ -470,6 +476,28 @@ def get_model_settings(self, model):
                     {settings_str}
                     """
 
+    @available
+    def filter_settings_by_engine(self, settings, engine):
+        filtered_settings = {}
+
+        if engine not in IGNORED_SETTINGS:
```
BentsiLeviav, line 483: The `IGNORED_SETTINGS` dict has only 2 keys (`S3` and `Memory`), therefore, any engine provided to this function will return an empty dict (`filtered_settings`). 

In case we don't have any "black list settings" for this engine, we should return the settings as is, no?

BentsiLeviav: We can remove this if clause, no? because if the engine is not in `IGNORED_SETTINGS`, we will iterate over an empty list and return the populated `filtered_settings` with the original `settings` values

shachibista: You're right 👍

BentsiLeviav (approved):

PR #370, opened 2024-10-21
feat: skip lightweight delete when there is no data to delete

Magicbeanbuyer:
## Summary

ClickHouse's lightweight delete function, which utilizes the mutation `ALTER TABLE table UPDATE _row_exists = 0 WHERE condition`, indiscriminately mutates all data parts of a given table. This occurs regardless of whether the data parts contain rows that need to be deleted or not, as detailed in this [issue](https://github.com/ClickHouse/ClickHouse/issues/59409).

Furthermore, when lightweight delete operations and regular merges are performed on the same table concurrently, the lightweight delete operation is forced to wait for the merge to complete before it can proceed. This leads to unnecessary delays when there are no rows to delete, rendering the `DELETE FROM...` operation wasteful.

Therefore, I propose implementing a preliminary check to determine if data in the table meets the deletion criteria before initiating a lightweight delete. If no matching data is found, the delete operation should be skipped, optimizing resource usage and improving overall efficiency.

## Checklist

### DIFF
--- a/dbt/include/clickhouse/macros/materializations/incremental/incremental.sql
+++ b/dbt/include/clickhouse/macros/materializations/incremental/incremental.sql
@@ -220,23 +220,44 @@
       {% endcall %}
     {% endif %}
 
-    {% call statement('delete_existing_data') %}
-      {% if is_distributed %}
-          {% set existing_local = existing_relation.incorporate(path={"identifier": this.identifier + local_suffix, "schema": local_db_prefix + this.schema}) if existing_relation is not none else none %}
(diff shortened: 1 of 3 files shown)


BentsiLeviav:
Hi again

Before reviewing your PR, it is required to add a short description with a link to this PR the the changelog (please keep the current format we have in the Changelog file).

BentsiLeviav (changes requested):
I think it would be best to put this feature behind a flag/setting because it adds a few more queries on top of the regular implementation. Additionally, there are cases where customers know (by definition or other means) that their data doesn't have empty increments, so disabling these unnecessary queries might be beneficial for them.

PR #384, opened 2024-11-14
fix: create materialized view on cluster

flytrap:
## materialized view

create materialized view need ON CLUSTER keyword


BentsiLeviav (approved):

BentsiLeviav:
This was reverted in #416 because it caused a few tests to fail. 
A workaround for this can be found here #328  , and a more holistic discussion about the issue can be found here #417 .

PR #394, opened 2024-11-28
Support of insert_overwrite in cluster setup

pheepa:
## Summary
Support of insert_overwrite in cluster setup with incremental and distributed_incremental materializations

## In details
1) In case is_distributed is True it creates distributed version of new data, gets partitions from cluster(system.parts) and replace in local existing table
2) The intermediate_relation entity was deleted as unnecessary because it is a incremental_legacy remnant. 
3) The existing test for insert_overwrite was renamed and made deterministic  

## Checklist


pheepa:
@BentsiLeviav Hi!
How does regular reviewer assignment work in this project?
What should be done to finish this?

BentsiLeviav (changes requested):
Generally LGTM.

One note - using `cluster()` ensures you query all nodes, but nodes may still be out of sync due to replication lag.
This means `system.parts` might not report all new partitions immediately.
@cwurm in #418 suggested adding `SYSTEM SYNC REPLICA` before querying `system.parts` to ensure that all data is visible before replacing partitions. Could you take care of this?

pheepa:
@BentsiLeviav 
Hi, thanks for the review! That’s a great point—I can take care of that.

However, I see another potential approach here. Wouldn't it be better to ensure all inserts in the cluster use insert_quorum=<number_of_nodes>? Right now, it's set to auto ([link](https://github.com/ClickHouse/dbt-clickhouse/blob/main/dbt/adapters/clickhouse/dbclient.py#L85)), which might cause issues when using a load balancer that routes queries to different replicas.

A possible problematic scenario:

1) You run model1 on one replica.
2) Then you run the child model2 on another replica.
The data might not be available due to replication lag.

But if one replica goes down, it could lead to a full system outage if insert_quorum is set too strictly.

That said, this might be too broad a change for an unrelated PR and should probably be discussed separately. What do you think?

BentsiLeviav (approved):

PR #403, opened 2025-01-06
chore(deps): upgrade dbt-core to ~=1.9.0

canbekley:
Bump dbt-core version to 1.9.x. #380 
Requires dbt-adapters >= 1.10.0

Issue #398, opened 2024-12-04
Add 'microbatch' incremental strategy


canbekley:
To better handle larger incremental time-series data processing, I would like to sugget adding support for the [microbatch](https://docs.getdbt.com/docs/build/incremental-microbatch) incremental_strategy. This feature requires dbt-core >= 1.9.0.

Here are some example implementations by other adapters:
- dbt-bigquery: https://github.com/dbt-labs/dbt-bigquery/pull/1334
- dbt-postgres: https://github.com/dbt-labs/dbt-postgres/pull/146
- dbt-snowflake: https://github.com/dbt-labs/dbt-snowflake/pull/1179

barberscott:
Parallel batch execution would be *huge* for some of the work we are doing in Clickhouse Cloud to allow us to distribute large incremental full-refresh builds across multiple nodes.

PR #404, opened 2025-01-06
feat: microbatch strategy

canbekley:
Adds support for the incremental "microbatch" strategy #398, that comes with dbt-core 1.9.0 #403
Currently only works for non-distributed incremental model, as dbt-core has hard-coded [config.materialization = 'incremental'](https://github.com/dbt-labs/dbt-core/blob/0bf38ce294e21c6e4056ed36ba8f35538c10604a/core/dbt/task/run.py#L665) into a microbatching condition. Maybe this can be patched by the adapater though?

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ### Release [x.x.x]
 ### Improvements
 * Add support for "microbatch" incremental strategy.
-* Ignores incompatible settings based on the configured Engine.
 
 ### Release [1.8.9], 2025-02-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### Release [x.x.x]
 ### Improvements
 * Add support for "microbatch" incremental strategy.
+* Ignores incompatible settings based on the configured Engine.
 
 ### Release [1.8.9], 2025-02-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Release [x.x.x]
+### Improvements
+* Add support for "microbatch" incremental strategy.
+
 ### Release [1.8.9], 2025-02-16
 
 #### Improvements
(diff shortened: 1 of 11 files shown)


canbekley:
@BentsiLeviav may I ask for a review on this? it would be an important addition for our pipelines

BentsiLeviav:
Hi @canbekley 

I'll review this in the next following days.
This feature depends on #403 , so feel free to answer there once you have time.
In addition, could you please add an indication of this feature to the README.md file? an explanation of how this was implemented would be appreciated

pheepa:
Hi, that's a great feature - thanks for implementing it!  

I don't quite understand the logic behind requiring `partition_by` for this method if microbatching is handled via a series of `delete_insert` strategy operations.  

Wouldn't it make more sense to either remove this restriction or implement it using `insert_overwrite`, similar to how it's done in dbt-bigquery? [Reference](https://github.com/dbt-labs/dbt-bigquery/blob/main/dbt/include/bigquery/macros/materializations/incremental_strategy/microbatch.sql#L23).  

In Snowflake, `delete_insert` are used without requiring `partition_by`: [Reference](https://github.com/dbt-labs/dbt-snowflake/blob/main/dbt/include/snowflake/macros/materializations/merge.sql#L53).  

Since ClickHouse struggles when handling a large number of complex mutations (my experience), I assume it would be better to go with `insert_overwrite`.

canbekley:
@pheepa I don't see where the `microbatch` strategy is requiring `partition_by`. You can also check the integration test model, which doesn't have a `partition_by` configuration, but a required `unique_key` instead. I have refactored some of the existing incremental strategy validations by moving them to a adapter method, including some of the `insert_overwrite` validations. Maybe that part was confusing?

I opted for `delete_insert` here because I think it is more flexibel, in that it doesn't require a partitioned table, and we can have arbitrary batch sizes independent of existing partition keys.

pheepa:
@canbekley I'm sorry, the refactoring of the incremental strategy validation confused me a little bit. Now it looks good to me.
Let's see how delete_insert goes

canbekley:
hi @BentsiLeviav, I've added documentation and fixed an issue with previous python versions. could you have a second look?

canbekley:
@BentsiLeviav Do you maybe have an idea how to circumvent this check in [MicrobatchModelRunner.is_incremental()](https://github.com/dbt-labs/dbt-core/blob/0bf38ce294e21c6e4056ed36ba8f35538c10604a/core/dbt/task/run.py#L665), which currently prevents `distributed_incremental` models being run in microbatches?

maowerner:
Hi, is there any update on this? I believe this feature is waiting for review :)

PR #407, opened 2025-01-13
fix: replicated database materializations

canbekley:
## Summary
Previously, table and view materializations wouldn't work with a `Replicated` database configured, as they would run `ON CLUSTER` statements for some of the ddl's. This is not allowed in replicated databases and would lead to the following error:
```
DB::Exception: It's not initial query. ON CLUSTER is not allowed for Replicated database. (INCORRECT_QUERY) (version 24.11.1.2557 (official build)
```
This mr fixes these issues by adding conditions for on cluster statements with replicated databases. 

## Checklist

### DIFF
--- a/dbt/include/clickhouse/macros/adapters.sql
+++ b/dbt/include/clickhouse/macros/adapters.sql
@@ -59,14 +59,13 @@
 
 {% macro clickhouse__drop_relation(relation, obj_type='table') -%}
   {% call statement('drop_relation', auto_begin=False) -%}
-    {# drop relation on cluster by default if cluster is set #}
-    drop {{ obj_type }} if exists {{ relation }} {{ on_cluster_clause(relation.without_identifier(), True)}}
+    drop {{ obj_type }} if exists {{ relation }} {{ on_cluster_clause(relation, True)}}
   {%- endcall %}
 {% endmacro %}
 
 {% macro clickhouse__rename_relation(from_relation, to_relation, obj_type='table') -%}
   {% call statement('drop_relation') %}
-    drop {{ obj_type }} if exists {{ to_relation }} {{ on_cluster_clause(to_relation.without_identifier())}}
+    drop {{ obj_type }} if exists {{ to_relation }} {{ on_cluster_clause(to_relation)}}
   {% endcall %}
   {% call statement('rename_relation') %}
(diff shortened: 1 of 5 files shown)


dbt/include/clickhouse/macros/adapters.sql
```diff
@@ -59,14 +59,13 @@
 
 {% macro clickhouse__drop_relation(relation, obj_type='table') -%}
   {% call statement('drop_relation', auto_begin=False) -%}
-    {# drop relation on cluster by default if cluster is set #}
-    drop {{ obj_type }} if exists {{ relation }} {{ on_cluster_clause(relation.without_identifier(), True)}}
+    drop {{ obj_type }} if exists {{ relation }} {{ on_cluster_clause(relation, True)}}
```
BentsiLeviav, line 62: Could you elaborate more on why to include back the identifier?

dbt/include/clickhouse/macros/adapters.sql
```diff
@@ -59,14 +59,13 @@
...
   {%- endcall %}
 {% endmacro %}
 
 {% macro clickhouse__rename_relation(from_relation, to_relation, obj_type='table') -%}
   {% call statement('drop_relation') %}
-    drop {{ obj_type }} if exists {{ to_relation }} {{ on_cluster_clause(to_relation.without_identifier())}}
+    drop {{ obj_type }} if exists {{ to_relation }} {{ on_cluster_clause(to_relation)}}
```
BentsiLeviav, line 68: Also here

BentsiLeviav (changes requested):

dbt/include/clickhouse/macros/adapters.sql
canbekley: these macros are used to drop/rename tables and other relation types. If the relation doesn't contain an identifier, the [ClickHouseRelation.should_on_cluster](https://github.com/ClickHouse/dbt-clickhouse/blob/587bd248d2892ab3cb7092b866891efe5a8d92cb/dbt/adapters/clickhouse/relation.py#L73) property will always return True, skipping a check on `can_on_cluster`. this was causing issues for tables in replicated databases, as they don't support `on cluster` statements.

BentsiLeviav (approved):

Issue #336, opened 2024-08-14
Question: How to create non-clustered materialized views and clustered tables


aaronli94:
I use the following code to create a materialized view to write to a table with ReplicatedReplacingMergeTree engine:
```
{{ config(
       materialized="materialized_view",
       engine="ReplicatedReplacingMergeTree('/clickhouse/tables/kafka/mysql_test_users/{uuid}','{replica}',_ts_ms,_deleted)",
       order_by="(id)"
) }}

SELECT
    JSONExtractString(_key, 'id') AS id,
    if(op = 'd',JSONExtractString(before,'name'),JSONExtractString(after,'name')) AS name,
	toInt32(if(op = 'd',JSONExtractString(before,'age'),JSONExtractString(after,'age'))) AS age,
	if(op = 'd',toDateTime64(JSONExtractString(before,'created_at'),3),
	    toDateTime64(JSONExtractString(after,'created_at'),3)) AS created_at,
	if(op = 'd',toDateTime64(JSONExtractString(before,'updated_at'),3),
	    toDateTime64(JSONExtractString(after,'updated_at'),3)) AS updated_at,
	if(op = 'd',1,0) as _deleted,
    ts_ms as _ts_ms
FROM
    {{ source('kafka', 'mysql_test_users_json') }}
```
This will be compiled into a view creation query with on cluster clause, however, I do not want to create a view in the cluster, but the engine of the target table needs on cluster clause, is there any way to create non-clustered materialized views and clustered tables?

BentsiLeviav:
As part of the 1.9 release, we changed the behavior of the `on cluster` clause. It is now possible to disable the on-cluster clause for specific models. 

Having said that, I agree with @reven404 that such behavior might lead to inconsistent DDL states across the cluster.

Issue #327, opened 2024-07-26
Table with Null engine not created on cluster


benjamin-awd:
### Describe the bug
Tables with null engines are not created using the `ON CLUSTER` clause

### Steps to reproduce
1. Attempt to create a relation with a null engine e.g.
```sql
{{
    config(
        engine="Null",
        materialized="materialized_view",
        schema="default",
        cluster="default"
    )
}}

select * from {{ ref("stg_testing") }}
```

2. Materialized view + target table is only created on a single server, and not across the entire cluster

### Expected behaviour
There should be an option to allow a null engine table to be created on cluster

### Code examples, such as models or profile settings
Also tried defining the cluster via profiles.yml, but this makes no difference

benjamin-awd:
I figure the quick fix for this is to update the logic in `get_on_cluster()`: 

```diff
    @classmethod
    def get_on_cluster(
        cls: Type[Self], cluster: str = '', materialized: str = '', engine: str = ''
    ) -> bool:
        if cluster.strip():
            return (
                materialized in ('view', 'dictionary')
                or 'distributed' in materialized
                or 'Replicated' in engine
+               or engine in ('Null')
            )

        else:
            return False
```

benjamin-awd:
Another relevant issue: https://github.com/ClickHouse/dbt-clickhouse/issues/237

emirkmo:
The current `on_cluster` macro is really overloaded. It is a conditional macro that checks for things other than `cluster`.

For example, I wanted to use it in a `post-hook` for configuring Row Policies but basically couldn't. 

Perhaps instead of adding more and more cases to the if statement, we can refactor it to work by default when called, and fix the problems from its misuse in other macros?

Or we should leave the current one as is since it is used everywhere, but make a `real_on_cluster_macro` that does not do any checking of the relation, just returns the `cluster` correctly formatted as `'ON CLUSTER <cluster>'` if cluster exists... :D . The current one can use the `real_on_cluster_macro` under the hood. It should really be called `conditional_on_cluster_macro` since it conditionally returns on_cluster if some conditions not specified in the README are met.

This is a sincere suggestion, what do you think @benjamin-awd? Then you could use it with any type of Engine etc. and the conditional one can be used by those implementations that need more logic such as distributed/Replicated materializations..

benjamin-awd:
[...]

Issue #237, opened 2024-01-12
engine = "Set" not created on cluster mode


ikeniborn:
{{ 
    config(
    enabled = true
    ,full_refresh = false
    ,schema = 'set'
    ,tags = ["discreteness_hour","model_token"]
    ,engine = "Set"
    ,materialized = "incremental"
    ,incremental_strategy = 'append'
    ,settings = { 'persistent': 1 }
    ,pre_hook = [
      "{{truncate_table_cluster(this)}}"
    ]
  ) 
}}

select 
  t0.token_id as token_id
from
  {{ref("snap_dim_token")}} t0
where 
  empty(main_token_id)
  and token_coinmarketcap_is_active
  and token_category_id not in (
      'e5e3fd01-394b-9a81-296b-75d5a7f4c1a2' /* STABLECOIN */
    ) 
  -- and token_id not in {{ref("set_dim_token_wrapped")}} 
  and token_name not like '%WRAP%'
  and token_id not in (
    '69c6b514-4293-1c8f-391b-40b376391c84' /* LIDO STAKED ETH [STETH]*/ 
  )
order by 
  t0.updated_dttm desc
limit 1 by 
  t0.token_id

in log 

[0m12:58:08.843851 [debug] [Thread-1 (]: dbt_clickhouse adapter: On model.clickhouse.set_dim_token_active: /* {"app": "dbt", "dbt_version": "1.7.4", "profile_name": "clickhouse", "target_name": "*****", "node_id": "model.clickhouse.set_dim_token_active"} */

            

    
        create table `set`.`set_dim_token_active`
        
  
        
  engine = Set
        
        
        
        SETTINGS  persistent=1

            empty
          as (
            

select 
  t0.token_id as token_id
from
  `snapshot`.`snap_dim_token` t0
where 
  empty(main_token_id)
  and token_coinmarketcap_is_active
  and token_category_id not in (
      'e5e3fd01-394b-9a81-296b-75d5a7f4c1a2' /* STABLECOIN */
    ) 
  -- and token_id not in `set`.`set_dim_token_wrapped` 
  and token_name not like '%WRAP%'
  and token_id not in (
    '69c6b514-4293-1c8f-391b-40b376391c84' /* LIDO STAKED ETH [STETH]*/ 
  )
order by 
  t0.updated_dttm desc
limit 1 by 
  t0.token_id
          )
        ...
Core:

installed: 1.7.4
latest: 1.7.4 - Up to date!
[...]

Issue #225, opened 2023-12-19
Generic test materializing not using ON CLUSTER


dstsimokha:
### Describe the bug

Generic tests with `store_failures = true` materializing **not** using ON CLUSTER clause and provided in profiles.yml value

### Expected behaviour

Generic tests with `store_failures = true` materializing using ON CLUSTER clause and provided in profiles.yml value

#### Environment
* dbt version: 1.7.3
* dbt-clickhouse version: 1.7.0
* clickhouse-driver version (if using native): 0.2.6
* clickhouse-connect version (if using http): 0.6.22
* Python version: 3.11.6
* Operating system: macos 14.2

YanisPlevokas:
@BentsiLeviav I suppose that in this case problem is not in get_on_cluster method - on the dbt-clickhouse version 1.7.7 and dbt.core 1.7.18, python version 3.8.10 I tried to change all get_on_cluster methods to return True - and still tables for test were created without 'ON CLUSTER <cluster_name>' block.
The problem (as I understand) in this line https://github.com/ClickHouse/dbt-clickhouse/blob/main/dbt/include/clickhouse/macros/materializations/table.sql#L132, relation.should_on_cluster is False for the test table relation, because in this lines https://github.com/ClickHouse/dbt-clickhouse/blob/main/dbt/adapters/clickhouse/relation.py#L71-L77 it returns self.can_on_cluster, that is defined as False here https://github.com/ClickHouse/dbt-clickhouse/blob/main/dbt/adapters/clickhouse/relation.py#L46 by default. Therefore, for the test relation we don't reassign can_on_cluster before using it, which leads to the problem of creating test tables without 'on cluster' block.

PR #447, opened 2025-04-25
Fix on cluster clause behaviour

### LINKED ISSUE #417: Refactor `get_on_cluster` logic to support flexible cluster configuration
## Summary
There have been multiple issues and PRs related to incorrect materialization and engine selection when using ON CLUSTER (#237,  #225 #336 #327  #328 #384 #307) . The current logic in get_on_cluster does not consistently apply the ON CLUSTER clause, leading to unexpected behavior when defining models in dbt.

This issue aims to consolidate the discussions and propose a refactor that provides a clear and flexible approach for handling cluster settings.

## Current Problems
* Inconsistent behavior: Some materialized views or tables do not get created across the cluster when `cluster` is defined.
* Overloaded logic: `get_on_cluster` has a complex if-statement, making it difficult to extend or use in different macros.
* Lack of user control: Users expect that if they provide `cluster`, models will be created on the cluster, but this is not always the case.
* Workarounds instead of solutions: PRs like #328 introduced `force_on_cluster`, which adds `ON CLUSTER` unconditionally, but a more systematic approach is needed.

## Proposed Solution:
Refactor `get_on_cluster` to support clear configuration options, ensuring predictable behavior based on user preferences.

**Configuration Options:**
1. **Always create models on the cluster:** If a user defines a cluster, and this materialization/engine is suitable for a cluster, all models should be created with `ON CLUSTER`.
2. **Never create models on the cluster:** Users can opt-out entirely, even if `cluster` is defined in their profile.
3. **Create all models on the cluster, except for specific ones:** A user can specify exceptions, ensuring most models are clustered while some remain local.
4. **Do not create models on the cluster, except for specific ones:** A user can disable clustering globally but enable it for selected models.
5. **Automatically apply `ON CLUSTER` for required cases:** Certain models, like `Distributed` and `Replicated` tables, should always use `ON CLUSTER`.

These options should be configurable in profiles.yml, ensuring a centralized approach to managing cluster settings.

BentsiLeviav:
## Summary
This PR refactors the logic around how the ON CLUSTER clause is applied during model creation in the dbt ClickHouse adapter. The motivation stems from ongoing confusion and inconsistency in how ON CLUSTER is handled, as documented in issues like #237, #225, #336, #327, #328, #384, and #307.

Close #417 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,12 @@
 ### Improvements
 * Upgrade `dbt-core` to version `1.9` and `dbt-adapters` to `>=1.10` ([#403](https://github.com/ClickHouse/dbt-clickhouse/pull/403))
 * Previously, delete_insert would fall back to legacy silently. Now it raises an error if LWD is not enabled.
+* Refactored the logic for applying the `ON CLUSTER` clause during model creation.
+    - Previously, `ON CLUSTER` was inconsistently applied based on engine type and materialization, often leading to confusion and unexpected results.
+    - Now, if a `cluster` is defined in the profile and the engine is **not** `Replicated`, the `ON CLUSTER` clause will **always** be added by default.
+    - A new model-level config `disable_on_cluster: true` has been introduced to explicitly opt out of this behavior.
+    - ⚠️ **Breaking Change**: This modifies the previous behavior. Users relying on the old implicit logic should review and update their models accordingly.
+
 
 ### Release [1.8.9], 2025-02-16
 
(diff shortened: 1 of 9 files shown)


genzgd (approved):

emirkmo:
May I ask how we are supposed to get ON CLUSTER to work with ReplicatedMergeTree tables now?

benjamin-awd:
Unfortunately as @emirkmo mentioned, this does not seem to allow ON CLUSTER with ReplicatedMergeTree tables.

PR #471, opened 2025-06-07
chore: Substitute deprecated pkg_resources

### LINKED ISSUE #470: Use importlib instead of pkg_resources
`pkg_resources` is deprecated ([see manual](https://setuptools.pypa.io/en/latest/pkg_resources.html)). It is only used to [discover the clickhouse-driver package version](https://github.com/ClickHouse/dbt-clickhouse/blob/6177aeba2a4d40fff77849b40998538230f5d102/dbt/adapters/clickhouse/nativeclient.py#L15). It should be easy to replace this with `importlib.metadata` [version](https://docs.python.org/3.13/library/importlib.metadata.html#distribution-versions) lookup. This is available since Python 3.8. [dbt requires Python 3.9+](https://docs.getdbt.com/faqs/Core/install-python-compatibility).

enqueue:
## Summary

Substitute use of pkg_resource API with importlib.metadata

Fixes #470

### DIFF
--- a/dbt/adapters/clickhouse/nativeclient.py
+++ b/dbt/adapters/clickhouse/nativeclient.py
@@ -1,7 +1,7 @@
 from typing import List
 
 import clickhouse_driver
-import pkg_resources
+from importlib.metadata import version, PackageNotFoundError
 from clickhouse_driver.errors import NetworkError, SocketTimeoutError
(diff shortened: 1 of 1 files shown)


morsapaes (approved):
Thanks, @enqueue!

PR #502, opened 2025-09-03
Fix CH version selection. Adds timeouts.

koletzilla:
## Summary
- Now all tests will fail if one of them takes longer than 120 seconds
- Selection of CH version is now fixed.
- latest version (25.8) will be ignored until https://github.com/ClickHouse/ClickHouse/issues/86434 is fixed. We could add a workaround here, but as it will be released soon, we decided just to skip it.
- No need for custom 22_3 settings anymore.


.github/workflows/test_matrix.yml
```diff
@@ -30,20 +32,28 @@ jobs:
           - '3.12'
         clickhouse-version:
           - '23.8'
-          - '24.1'
-          - '24.2'
           - '24.3'
-          - latest
+          - '24.8'
+          - '25.3'
+          - '25.7'  # Skipping 25.8+ until https://github.com/ClickHouse/ClickHouse/issues/86434 is fixed.
```
koletzilla, line 38: Removed 24.1 and 24.2 in favour of LTS versions

tests/integration/adapter/clickhouse/test_clickhouse_table_ttl.py
```diff
@@ -7,7 +7,13 @@
 from dbt.tests.adapter.basic.test_base import BaseSimpleMaterializations
 from dbt.tests.util import relation_from_name, run_dbt
 
+from tests.integration.adapter.helpers import below_version
 
+
+@pytest.mark.skipif(
+    below_version(24, 3),
+    reason='Pending to fix. Syntax error in 23.8: Code: 62. DB::Exception: Syntax error: failed at position 501 (\'SETTINGS\') (line 20, col 21): SETTINGS  allow_nullable_key=1,  replicated_deduplication_window=0',
+)
```
koletzilla, line 16: I'd like to fix all these 23.8 errors now, but as these tests have been already failing for a while, I'd rather leave them as they are right now and move to other more important issues

tests/integration/adapter/constraints/fixtures_constraints.py
```diff
@@ -285,6 +285,6 @@
 
 select
   101::Int32 as id,
-  timestamp('2025-04-16') as ts,
+  UTCtimestamp() as ts,
```
koletzilla, line 288: `timestamp` was added in 23.9. This caused a test to fail in 23.8

tests/integration/adapter/helpers.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+
+def below_version(major: int, minor: int = 0) -> bool:
+    current_version = os.environ.get('DBT_CH_TEST_CLOUD', '0.0')
+    actual_major, actual_minor = current_version.split('.')
+    return int(actual_major) < major or int(actual_major) == major and int(actual_minor) < minor
```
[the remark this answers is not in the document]
koletzilla: We are going to use major.minor versions. No need for other cases now

pytest.ini
```diff
@@ -1,9 +1,11 @@
 [pytest]
+log_cli = True
```
BentsiLeviav: Small note
Without this, pytest usually only shows log output if a test fails - are we sure we want to see all logs in all cases? Won't it overwhelm the page when we try to see the failed logs?

tests/integration/adapter/clickhouse/test_clickhouse_table_ttl.py
BentsiLeviav: Can we first create a dedicated issue for tracking all 23.8 errors?

tests/integration/adapter/helpers.py
```diff
@@ -0,0 +1,22 @@
...
+    >>> below_version(25, 7, _ch_test_version_value='25.7')
+    False
+    """
+    current_version = (
+        _ch_test_version_value
+        or os.environ.get('DBT_CH_TEST_CH_VERSION', '0.0')
+        or '0.0'  # Extra 0.0 to make Mypy happy
+    )
+    actual_major, actual_minor = current_version.split('.')
+    return int(actual_major) < major or int(actual_major) == major and int(actual_minor) < minor
```
BentsiLeviav: I know that `and` has higher precedence than `or`, but adding this might avoid confusion 
```suggestion
    return int(actual_major) < major or (int(actual_major) == major and int(actual_minor) < minor)
```

BentsiLeviav (approved):

tests/integration/adapter/clickhouse/test_clickhouse_table_ttl.py
koletzilla: Sounds good! Created it: https://github.com/ClickHouse/dbt-clickhouse/issues/503

pytest.ini
koletzilla: It was useful while debugging, but you're right: It's probably not going to be as useful in our day-to-day basics. Removing it

tests/integration/adapter/helpers.py
koletzilla: Sounds good!

Issue #510, opened 2025-09-10
Macro clickhouse__get_columns_in_relation sometimes is not filtering by schema


koletzilla:
### Describe the bug
Happened while working on https://github.com/ClickHouse/dbt-clickhouse/issues/493

clickhouse__get_columns_in_relation should always use the schema (the ClickHouse database) to filter when querying `system.columns` or it may get results from other tables in other databases.

The test `TestSnapshotCheckCols` was failing for us as the temp table `cc_name_snapshot__dbt_tmp` was already created in another database, so this query was returning twice the columns and so making the snapshot process fail.

Looks like in this case the culpit is this macro:
```jinja2
{% macro clickhouse__make_temp_relation(base_relation, suffix) %}
  {% set tmp_identifier = base_relation.identifier ~ suffix %}
  {% set tmp_relation = base_relation.incorporate(
                              path={"identifier": tmp_identifier, "schema": None}) -%}
  {% do return(tmp_relation) %}
{% endmacro %}
```
This is the macro that creates the temp table. Notice the `schema": None`. If instead of None we add `base_relation.schema`, the test passes.

### Steps to reproduce
1. Make sure an existing temp table exists
2. execute the test.

Also you can test it by adding `exceptions.raise_compiler_error` inside the `get_columns_in_relation`. We may need to add at least a `exceptions.warn` there to ensure this is not happening in other places.

```jinja2
{% macro clickhouse__get_columns_in_relation(relation) -%}
  {% call statement('get_columns', fetch_result=True) %}
    select name, type from system.columns where table = '{{ relation.identifier }}'
    {% if relation.schema %}
      and database = '{{ relation.schema }}'
    {% else %}
        {% do exceptions.raise_compiler_error("Relations should always come with a defined schema. Missing schema for " ~ relation.identifier) %}
    {% endif %}
    order by position
  {% endcall %}
  {{ return(sql_convert_columns_in_relation(load_result('get_columns').table)) }}
{% endmacro %}
```

### Expected behaviour
Queries to system tables to ask for table metadata should **ALWAYS** filter by database.

PR #511, opened 2025-09-10
Always include the schema when creating a temp relation

koletzilla:
## Summary
Fixes https://github.com/ClickHouse/dbt-clickhouse/issues/510

The new temp relation wasn't including the schema. That caused some queries like the one to `system.columns` to get incorrect values.

After fixing it I cannot find more situations where this is problematic.

This also should put us closer to the parallelisation of the cloud tests.

### DIFF
--- a/dbt/include/clickhouse/macros/adapters.sql
+++ b/dbt/include/clickhouse/macros/adapters.sql
@@ -52,7 +52,7 @@
     {% if relation.schema %}
       and database = '{{ relation.schema }}'
     {% else %}
-      {% do exceptions.raise_compiler_error("Relations should always come with a defined schema. Missing schema for " ~ relation.identifier) %}
+      {% do exceptions.warn("Relations should always come with a defined schema. Missing schema for " ~ relation.identifier) %}
     {% endif %}
     order by position
   {% endcall %}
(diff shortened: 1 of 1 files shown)


dbt/include/clickhouse/macros/adapters.sql
```diff
@@ -51,6 +51,8 @@
     select name, type from system.columns where table = '{{ relation.identifier }}'
     {% if relation.schema %}
       and database = '{{ relation.schema }}'
+    {% else %}
+        {% do exceptions.raise_compiler_error("Relations should always come with a defined schema. Missing schema for " ~ relation.identifier) %}
```
koletzilla: We can also add `warn` instead of `raise_compiler_error ` if we think this may impact some user. I don't have an strong opinion with any of this.

I feel more secure adding the error as in any situation this call should be done without filtering by the database, but we may also add warning to let users know that they should contact us with the details.

BentsiLeviav: Let's use warn to be on the safe side

koletzilla: Sounds good. Changed!

Issue #458, opened 2025-05-12
Fix ClickHouse cloud tests


BentsiLeviav:
### Describe the bug

Fix the broken GitHub workflow for dbt Cloud integration tests and add a nightly scheduled run to catch regressions.

koletzilla:
I have been working on it here and right now we have a set of tests failing. They're all related to `GRANT` operations like `test_incremental_grants.py`. Looks like the grant is not synced at the moment of the creation with all replicas, so if the GRANT check is done to one replica that has not received it, the test will fail.

Same thing happens with the `CREATE USER` right at the beginning. 

I have tried a few things:
- `select_sequential_consistency = 1 `worked to wait for data in tables but don't seem to affect here.
- in CH cloud there's no `ON CLUSTER` operation so `distributed_ddl_task_timeout` don't apply.
- I cannot do `SYSTEM SYNC DATABASE REPLICA default` as it seems it only works for replicated databases and in CH cloud `default` is `Shared`.
 
I have asked for more ways to solve this, but right now the only way to solve it is to:
1. Get the number of replicas by doing `SELECT count() from system.clusters WHERE database = 'default'`
2. Wait from Python code until `select count() from clusterAllReplicas('default', system.users)` where ...` (or `FROM clusterAllReplicas('default', system.grants)') is the same as the number of replicas we found before.

PR #512, opened 2025-09-11
Improve codebase to reintegrate cloud tests ASAP.

koletzilla:
## Summary

This PR adds back cloud tests so it will be executed on PRs again. I feel like they are now in a good enough shape to start executing them also in PRs. It also tests with the new CH Cloud cluster which is more stable than the other one.

Pending things:
- https://github.com/ClickHouse/dbt-clickhouse/issues/458: This issue should be closed a few days after we merge this PR, but lets wait a bit in case we found other problematic tests.
- https://github.com/ClickHouse/dbt-clickhouse/issues/532: We need to also test with other python versions. Testing with other versions is currently not the most problematic thing as we do test them in test_matrix, but it would be nice to also have these tests here. We may add these other python versions once we optimise the tests so they run faster or at least they can run concurrently.

---

> [!NOTE]
> Re-enables Cloud testing in CI and makes tests compatible with Cloud limitations.
> 
> - **CI**: Update `test_cloud.yml` to trigger on PRs to `main` and set a `timeout-minutes: 50` for `cloud_smt_tests`.
> - **Tests**: Conditionally skip distributed/replicated constraints suites (`TestDistributed*ContractColumnsEqual`) when `DBT_CH_TEST_CLOUD` is true, since replicated tables aren’t supported in Cloud; add `os` import to support skip logic.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 234d7495e62efd46e8c21fb8ffbdb1bb19770066. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/integration/adapter/constraints/test_constraints.py
+++ b/tests/integration/adapter/constraints/test_constraints.py
@@ -1,3 +1,5 @@
+import os
+
 import pytest
 from dbt.tests.util import get_manifest, run_dbt, run_dbt_and_capture, write_file
 
@@ -151,6 +153,10 @@ class TestViewContractColumnsEqual(ClickHouseContractColumnsEqual):
         }
 
 
+@pytest.mark.skipif(
+    os.environ.get('DBT_CH_TEST_CLOUD', '').lower() in ('1', 'true', 'yes'),
+    reason='Replicated is not supported for cloud',
+)
 class TestDistributedTableContractColumnsEqual(ClickHouseContractColumnsEqual):
     @pytest.fixture(scope="class")
     def models(self):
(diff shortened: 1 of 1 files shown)


tests/integration/adapter/constraints/test_constraints.py
```diff
@@ -180,6 +186,10 @@ def test_incremental_contract_wrong_column_names(self, project):
         contract_wrong_column_names(project)
 
 
+@pytest.mark.skipif(
+    os.environ.get('DBT_CH_TEST_CLOUD', '').lower() in ('1', 'true', 'yes'),
+    reason='Replicated is not supported for cloud',
+)
```
koletzilla, line 192: Those test were added in https://github.com/ClickHouse/dbt-clickhouse/pull/575 but they fail in cloud. The issue is that they test Distributed tables and we cannot test them for cloud. They raise this error:
```python
14:02:16    Compilation Error in model my_model_wrong_order (models/my_model_wrong_order.sql)
  To use distributed materialization cluster setting in dbt profile must be set
  
  > in macro materialization_distributed_table_clickhouse (macros/materializations/distributed_table.sql)
  > called by model my_model_wrong_order (models/my_model_wrong_order.sql)
```
Skipping them.

Issue #514, opened 2025-09-12
Regression in 1.9.3: Unit Tests With Fixtures Failing With "Temporary tables cannot be inside a database"


enqueue:
### Describe the bug

When running a unit test with fixtures using dbt-clickhouse version 1.9.3, the following error occurs:

> Code: 442. DB::Exception: Temporary tables cannot be inside a database.
> You should not specify a database for a temporary table.
> (BAD_DATABASE_FOR_TEMPORARY_TABLE) 

This error does _not_ occur using dbt-clickhouse version 1.9.2.

Query in 1.9.3:

```sql
create temporary table `dbname`.`testname__dbt_tmp` engine Memory -- end_of_sql
 as ( select * from ( with __dbt__cte__table_to_test as ...
```

Query in 1.9.2:

```sql
create temporary table `testname__dbt_tmp`  engine Memory  -- end_of_sql
as ( select * from (  with __dbt__cte__table_to_test as ...
```

### Steps to reproduce
1. Create a unit test with a CSV fixture
2. Run dbt test

### Expected behaviour

Test to be green ✅ 

### Code examples, such as models or profile settings

I could perhaps create a reproducer if it is really required, but it would take me some time. Perhaps you can already pinpoint the error without having a complete example project.

### dbt and/or ClickHouse server logs

ClickHouse error log:

```
<Error> executeQuery: Code: 442. DB::Exception: Temporary tables cannot be inside a database. You should not specify a database for a temporary table. (BAD_DATABASE_FOR_TEMPORARY_TABLE)
```

dbt log:

```
10:45:52    Runtime Error in unit_test_my_test (models/basic/my_test.yml)
  An error occurred during execution of unit test 'my_test'. There may be an error in the unit test definition: check the data types.
   Database Error
    Code: 442.
    DB::Exception: Temporary tables cannot be inside a database. You should not specify a database for a temporary table.. Stack trace:
```

#### Environment
* dbt version: 1.10.8
* dbt-clickhouse version: 1.9.3
* clickhouse-driver version: 0.2.9
* clickhouse-connect version: 0.8.17
* Python version: 3.13.7
* Operating system: Linux

#### ClickHouse server
* ClickHouse Server version: 25.8.1.5101
* ClickHouse Server non-default settings, if any: _none_
* `CREATE TABLE` statements for tables involved: n/a
[...]

PR #515, opened 2025-09-12
Ensure that temporary tables are not accessed with database clause

koletzilla:
Fixes https://github.com/ClickHouse/dbt-clickhouse/issues/514
- Makes sure temporary tables are created without database and that `clickhouse__get_columns_in_relation ` is able to query without database for this situation.
- Add missing tests. These tests were failing because of the error reported in https://github.com/ClickHouse/dbt-clickhouse/issues/514

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 ### Release [1.9.4], 2025-XX-XX
 
 #### Bugs
-* Fix Materialized View not dropped when a model's materialization is changed from materialized_view to view ([#516](https://github.com/ClickHouse/dbt-clickhouse/pull/516))
+* Fix Materialized View not dropped when a model's materialization is changed from materialized_view to view ([#516](https://github.com/ClickHouse/dbt-clickhouse/pull/516)).
+* Ensure that temporary tables are not accessed with database clause ([#515](https://github.com/ClickHouse/dbt-clickhouse/pull/515)).
 
 
 ### Release [1.9.3], 2025-09-08
@@ -26,6 +27,7 @@
 * when the last line of a model's SQL query is a comment (-- some comment) and the table's contract is enforced, the last parenthesis of the wrapping subquery ends up commented as well. Was fixed in ([#457](https://github.com/ClickHouse/dbt-clickhouse/pull/457))
 * Check for Shared database engine in can_exchange ([#460](https://github.com/ClickHouse/dbt-clickhouse/pull/460))
 * Tests were broken because of docker compose version `2.35` and fixed in ([#468](https://github.com/ClickHouse/dbt-clickhouse/pull/468))
+
 
 ### Release [1.9.1], 2025-04-28
 


dbt/include/clickhouse/macros/adapters.sql
```diff
@@ -50,7 +50,9 @@
   {% call statement('get_columns', fetch_result=True) %}
     select name, type from system.columns where table = '{{ relation.identifier }}'
     {% if relation.schema %}
-      and database = '{{ relation.schema }}'
+      {% if not relation.identifier.endswith("__dbt_tmp") %}
+        and database = '{{ relation.schema }}'
+      {% endif %}
```
koletzilla: To use this we need to ensure that `__dbt_tmp` is only used for CH's `temporary` tables. If that suffix is added to other kind of tables, we need to think on a different way (maybe just a different suffix) to identify CH's `temporary` tables

koletzilla: I have checked it deeply, I hope I didn't miss something: all `__dbt_tmp` tables are temporary ones, so we are safe adding that.

tests/integration/adapter/unit_testing/test_types.py
```diff
@@ -0,0 +1,30 @@
...
+            ["true", "true"],
+            ["DATE '2020-01-02'", "2020-01-02"],
+            [
+                "TIMESTAMP '2013-11-03 00:00:00'",
+                "2013-11-03 00:00:00",
+            ],  # overrided: no "-0" support
+            [
+                "'2023-10-29 01:30:00'::DateTime('UTC')",
+                "2013-11-03 00:00:00",
+            ],  # overrided: no "-0" support and timezones expressed differently
```
koletzilla: Needed to override these two entryies from original code to make it work. Original was:
```python
            ["TIMESTAMP '2013-11-03 00:00:00-0'", "2013-11-03 00:00:00-0"],
            ["TIMESTAMPTZ '2013-11-03 00:00:00-0'", "2013-11-03 00:00:00-0"],
```
I wasn't able to execute it with `-0` nor with `TIMESTAMPTZ`

BentsiLeviav:
Is the problem only with creation, or with querying as well?

If it is the former, we can try to change this
https://github.com/ClickHouse/dbt-clickhouse/blob/87bbe8ef0df6289018f01d8e489407f6cf5dd1bf/dbt/include/clickhouse/macros/materializations/table.sql#L196-L203

to be something like this

```
 {% if temporary -%}
        create temporary table {{ relation.name }}
        engine Memory
        {{ adapter.get_model_settings(model, 'Memory') }}
        as (
          {{ sql }}
        )
```

dbt/include/clickhouse/macros/materializations/unit.sql
```diff
@@ -0,0 +1,34 @@
...
+
+  {% set relations = [] %}
+
+  {% set expected_rows = config.get('expected_rows') %}
+  {% set expected_sql = config.get('expected_sql') %}
+  {% set tested_expected_column_names = expected_rows[0].keys() if (expected_rows | length ) > 0 else get_columns_in_query(sql) %} %}
+
+  {%- set target_relation = this.incorporate(type='table') -%}
+  {%- set temp_relation = make_temp_relation(target_relation, '__dbt_tmp')-%}
+  {%- set temp_relation = temp_relation.incorporate(is_temporary=True)-%} {# file overwrite to ensure this is marked as temporary #}
```
koletzilla, line 11: @BentsiLeviav tested this new way of detecting if a relation is 'is_temporary' and is looking great. The only problem is that the only way I have found to implement this is to actually copy-paste the `unit.sql` file. other option would have been to completely override the base `make_temp_relation ` from the adapter but I'm not sure if that's possible.

I have seen that DBT supports `make_temp_relation ` and `make_intermediate_relation`. Looks like we are using them in the same way but may we should be using `make_intermediate_relation` for the rest of situations and actually `make_temp_relation` for the real, `temporary` tables. What do you think? that would let us implement this without having to copy-paste the unit.sql file and will probably make things easier in the future


I also confirmed one of the things we commented: `__dbt_tmp` is used in other tables like in the `distributed_table.sql` file but in these cases, these tables are not created as `temporary`. 'unit.sql' is the only place I have seen where `get_create_table_as_sql` is called with the first arg as True.

BentsiLeviav: @koletzilla 

What do you think about copying this for now, and creating a separate issue detailing this problem, where we also mention that the `units.sql` override will be removed once we align all intermediate/temp relations?

koletzilla: Sounds good to me! I'm going to do it.

koletzilla: Added https://github.com/ClickHouse/dbt-clickhouse/issues/524

PR #516, opened 2025-09-15
Click house/fix #507 mv not dropped

### LINKED ISSUE #507: Fully drop MV on materialization change
### Describe the bug
When a dbt model is switched from a materialized view to a view and redeployed, the old materialized view is not dropped. Instead, only the MV’s target table is removed. This leaves the MV object in place, causing problems when inserting data, such as:

``` Failed to insert data into ClickHouse (will be retried). Missing materialized view data is possible due to ClickHouse deduplication: ClickHouse Error Code 48: Method write is not supported by storage View.``` 

### Steps to reproduce
1. Create an MV.
1. Redeploy it as a view.
3. The MV does not get dropped, and the error message starts coming in.

### Expected behaviour

Both the target table and the MV are supposed to be removed.

### PR DESCRIPTION
**Pull request by koletzilla**
## Summary
Fixes https://github.com/ClickHouse/dbt-clickhouse/issues/507

Pending to be completely tested

### CODE REVIEW

File: dbt/include/clickhouse/macros/materializations/materialized_view.sql [hunk 1]
```diff
@@ -61,31 +61,15 @@
...
-            where table_schema = '{{ existing_relation.schema }}'
-              and table_name like '%{{ target_relation.name }}%'
-              and table_type = 'VIEW'
-        {% endset %}
-
-    {% set tables_result = run_query(tables_query) %}
-    {% if tables_result is not none and tables_result.columns %}
-        {% set tables = tables_result.columns[0].values() %}
-        {{ log('Current mvs found in ClickHouse are: ' + tables | join(', ')) }}
-        {% set mv_names = [] %}
```
koletzilla, line 77: Moved this logic to a function below to reuse it.

dbt/include/clickhouse/macros/materializations/materialized_view.sql
```diff
@@ -203,6 +187,52 @@
   {% endfor %}
 {%- endmacro %}
 
+{% macro clickhouse__search_associated_mvs_to_target(relation_schema, relation_name, mv_suffixes)  -%}
+  {% set tables_query %}
+    select name
+    from system.tables
+    where engine = 'MaterializedView'
+      and extract(create_table_query, 'TO\\s+([^\\s(]+)') = '{{ relation_schema }}.{{ relation_name }}'
+  {% endset %}
```
koletzilla, line 196: One important change here: Instead of calling information_schema.tables we are now calling system.tables. That way we can exactly know which MVs are pointing to the table being removed.

dbt/include/clickhouse/macros/materializations/view.sql
```diff
@@ -14,6 +14,7 @@
     {{ log('Creating new relation ' + target_relation.name )}}
   {% else %}
     {{ log('Relation ' + target_relation.name + ' already exists, replacing it' )}}
+    {{ clickhouse__drop_associated_mv_if_it_was_automatically_created(target_relation) }}
```
koletzilla, line 17: This way we remove only the `_mv` MV if found. We don't automatically remove other MVs but we print them in the logs. Two issues here:
- Adding this line here solves the issue, but to be honest I don't know if we need to also add it in other places.
- If we confirm we are removing the table here 100% of the times, we can actually remove automatically all other MVs pointing to this table. For safety reasons I have decided not to remove all of them them now.

dbt/include/clickhouse/macros/materializations/materialized_view.sql
```diff
@@ -203,6 +187,52 @@
   {% endfor %}
 {%- endmacro %}
 
+{% macro clickhouse__search_associated_mvs_to_target(relation_schema, relation_name, mv_suffixes)  -%}
+  {% set tables_query %}
+    select name
+    from system.tables
+    where engine = 'MaterializedView'
+      and extract(create_table_query, 'TO\\s+([^\\s(]+)') = '{{ relation_schema }}.{{ relation_name }}'
```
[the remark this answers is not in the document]
koletzilla: Not going to apply changes: the SQL to define it comes already formated from CH, so I don't think this is actually fragile at all.

dbt/include/clickhouse/macros/materializations/materialized_view.sql
```diff
@@ -203,6 +187,52 @@
...
+{%- endmacro %}
+
+
+{% macro clickhouse__drop_associated_mv_if_it_was_automatically_created(target_relation)  -%}
+  {#-
+    Limitations of this logic:
+     - Only covers situations where 1 mv was created. Don't cover multiple mvs with different names.
+     - Only checks current relation's database.
+    We print logs in case we find other mvs in that database
+  -#}
```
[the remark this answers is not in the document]
koletzilla: We already have these concerns in mind for the MVs revamp

dbt/include/clickhouse/macros/materializations/materialized_view.sql
```diff
@@ -203,6 +187,52 @@
...
+  {% endfor %}
+  {{ log('Model mvs to replace ' + mv_names | string) }}
+
+  {% set tables_result = run_query(tables_query) %}
+  {% if tables_result is not none and tables_result.columns %}
+    {% set tables = tables_result.columns[0].values() %}
+    {{ log('Current mvs found in ClickHouse are: ' + tables | join(', ')) }}
+    {{ return((tables, mv_names,)) }}
+  {% else %}
+    {{ return((None, mv_names,)) }}
```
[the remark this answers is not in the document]
koletzilla: Return only expects one argument.

dbt/include/clickhouse/macros/materializations/view.sql
BentsiLeviav: Will we get into this else block if the previous materialization is a table?

koletzilla: I think in fact this is what is actually happening 😓 I mean, `existing_relation` just know that a model exists, and it expects the model to be always a table so it behaves in the same way for DBT's materialized view and tables.

In both situations, having a mv pointing to the new created view is a problem, so removing any possible existing materialized view following the naming convention `modelname`+`_mv` is appropiate.

BentsiLeviav (approved):

PR #517, opened 2025-09-16
Remove support for non-maintained CH versions. Add again support for 25.8 (latest)

koletzilla:
## Checklist

### DIFF
--- a/.github/workflows/test_matrix.yml
+++ b/.github/workflows/test_matrix.yml
@@ -30,12 +30,11 @@ jobs:
           - '3.10'
           - '3.11'
           - '3.12'
-        clickhouse-version:
-          - '23.8'
-          - '24.3'
-          - '24.8'
+        clickhouse-version:  # Testing ClickHouse versions with active security support https://github.com/ClickHouse/ClickHouse/blob/master/SECURITY.md
           - '25.3'
-          - '25.7'  # Skipping 25.8+ until https://github.com/ClickHouse/ClickHouse/issues/86434 is fixed.
+          - '25.6'
+          - '25.7'
+          - 'latest'
 
     steps:
       - name: Checkout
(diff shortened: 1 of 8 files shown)


tests/integration/adapter/helpers.py
```diff
@@ -20,3 +21,30 @@ def below_version(major: int, minor: int = 0, _ch_test_version_value: Optional[s
...
+
+
+retry_config = TypedDict('retry_config', {'max_retries': int, 'delay': float})
+DEFAULT_RETRY_CONFIG: retry_config = {
+    "max_retries": 20,
+    "delay": 0.5,
+}
+
+
+def retry_assert_until_true(
```
BentsiLeviav: The function name `retry_assert_until_true` implies it retries until something becomes True, which is not the case. Can we rename it? maybe something like `retry_until_assertion_passes` or `retry_on_assertion_error`

tests/integration/adapter/helpers.py
```diff
@@ -20,3 +21,30 @@ def below_version(major: int, minor: int = 0, _ch_test_version_value: Optional[s
...
+    func: Callable[[], Any],
+    max_retries: int = DEFAULT_RETRY_CONFIG["max_retries"],
+    delay: float = DEFAULT_RETRY_CONFIG["delay"],
+) -> Any:
+    last_error: Optional[AssertionError] = None
+    for attempt in range(max_retries + 1):  # +1 to include the initial attempt
```
BentsiLeviav, line 39: Can you clarify this +1?
If `max_retries` is 3, this for loop will run 4 times (0,1,2,3), no?

koletzilla: if `max_retries=1`, range(1) will just return [0], so it will actually not retry it. It will execute the code only once.
```python
max_retries=1
[attemp for attemp in range(max_retries)]
Out[8]: [0]
```
Same when it's 3, it will not retry 3 times, just 2 (0 for the actual execution, 1 and 2 the retries)
```python
max_retries=3
[attemp for attemp in range(max_retries)]
Out[10]: [0, 1, 2]
```
Added the +1 to add the last iteration.

tests/integration/adapter/helpers.py
koletzilla: Good point! changed to `retry_until_assertion_passes`

BentsiLeviav (approved):

PR #522, opened 2025-09-18
Fix: Improve LowCardinality validation for model contracts

### LINKED ISSUE #303: Could we specify column type, aggregation function and compression on materialized view?
### Describe the bug
As topic, Could we specify column type and compression on materialized view?

Following the below example table:
```sql
CREATE TABLE IF NOT EXISTS db.mv_table ON CLUSTER '{cluster}'
(
    dt_day DateTime CODEC(Delta, ZSTD(1)),
    team_id LowCardinality(String) CODEC(ZSTD(1)),
    "le_accumulate" SimpleAggregateFunction(sum, Float64),
.......
)ENGINE = ReplicatedAggregatingMergeTree('/clickhouse/clusters/{cluster}/databases/{database}/tables/{table}/{shard}', '{replica}')
PARTITION BY (toYYYYMMDD(dt_day))
```

How do we create *materialized_view* and specify column type as like ` LowCardinality(String)` or compression as `CODEC(ZSTD(1))` as the same time? Additionally, how to we specify `SimpleAggregateFunction(sum, Float64)` during create *materialized_view*

If have any ideas can solve this, I will be appreciated it.

jorgeparavicini:
## Summary
This pull request resolves #303 by allowing the definition of columns as `LowCardinality` in the contracts. Before, defining a column of type `String` to be of type `LowCardinality(String)` caused an error if the constraints were enforced; otherwise, the `LowCardinality` was just ignored. 

Additionally, this PR allows defining special types that ignore the basic string comparison in the column constraints. 
At the moment, only the `SimpleAggregateFunction` is marked as a special type

## Checklist

### DIFF
--- a/dbt/include/clickhouse/macros/column_spec_ddl.sql
+++ b/dbt/include/clickhouse/macros/column_spec_ddl.sql
@@ -34,15 +34,27 @@
     {%- set yaml_data_type = yaml_col[0]['data_type'] -%}
     {%- set sql_data_type = sql_col['data_type'] -%}
 
-    {% if yaml_data_type.lower().startswith('lowcardinality') -%}
-      {#-- If contract is LowCardinality, extract the inner type for comparison --#}
-      {%- set yaml_data_type = yaml_data_type[15:-1] -%}
-    {%- endif -%}
+    {%- set ns = namespace(is_special_type=false) -%}
 
-    {%- if sql_data_type != yaml_data_type -%}
-      {#-- Column data types don't match #}
-      {%- do exceptions.raise_contract_error(yaml_columns, sql_columns) -%}
-    {%- endif -%}
+    {%- set special_types = ['SimpleAggregateFunction'] -%}
+    {%- for special_type in special_types -%}
+      {%- if yaml_data_type.startswith(special_type) -%}
+        {%- set ns.is_special_type = true -%}
+        {%- break -%}
+      {%- endif -%}
+    {%- endfor -%}
+
+    {% if not ns.is_special_type %}
+      {%- if yaml_data_type.startswith('LowCardinality') -%}
+        {#-- If contract is LowCardinality, extract the inner type for comparison --#}
+        {%- set yaml_data_type = yaml_data_type[15:-1] -%}
+      {%- endif -%}
+
+      {%- if sql_data_type != yaml_data_type -%}
+        {#-- Column data types don't match #}
+        {%- do exceptions.raise_contract_error(yaml_columns, sql_columns) -%}
+      {%- endif -%}
+    {% endif %} 
   {%- endfor -%}
 
 {% endmacro %}
--- a/dbt/include/clickhouse/macros/column_spec_ddl.sql
+++ b/dbt/include/clickhouse/macros/column_spec_ddl.sql
@@ -30,7 +30,16 @@
       {#-- Column with name not found in yaml #}
       {%- do exceptions.raise_contract_error(yaml_columns, sql_columns) -%}
     {%- endif -%}
-    {%- if sql_col['data_type'] != yaml_col[0]['data_type'] -%}
+
+    {%- set yaml_data_type = yaml_col[0]['data_type'] -%}
+    {%- set sql_data_type = sql_col['data_type'] -%}
+
+    {% if yaml_data_type.lower().startswith('lowcardinality') -%}
+      {#-- If contract is LowCardinality, extract the inner type for comparison --#}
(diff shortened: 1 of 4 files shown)


dbt/include/clickhouse/macros/column_spec_ddl.sql
```diff
@@ -30,10 +30,31 @@
...
+      {%- if yaml_data_type.startswith('LowCardinality') -%}
+        {#-- If contract is LowCardinality, extract the inner type for comparison --#}
+        {%- set yaml_data_type = yaml_data_type[15:-1] -%}
+      {%- endif -%}
+
+      {%- if sql_data_type != yaml_data_type -%}
+        {#-- Column data types don't match #}
+        {%- do exceptions.raise_contract_error(yaml_columns, sql_columns) -%}
+      {%- endif -%}
+    {% endif %}
```
koletzilla: the `{%- if sql_data_type != yaml_data_type -%}` is inside the `not ns.is_special_type`, so if the data type is "is_special_type", it don't gets validated.

As we are introducing some easy validations also for SimpleAggregateFunction here, what do you think if we also remove the `SimpleAggregateFunction` and get the inside type validated? I have modify your changes to this so:
- if `SimpleAggregateFunction`, it gets the second argument.
- same logic with `LowCardinality`
- then compare `sql_data_type != ns.yaml_data_type`

```jinja2
    {%- set ns = namespace(yaml_data_type=yaml_col[0]['data_type']) -%}
    {%- set sql_data_type = sql_col['data_type'] -%}

    {%- set special_types = ['SimpleAggregateFunction'] -%}
    {%- for special_type in special_types -%}
      {%- if ns.yaml_data_type.startswith(special_type) -%}
        {%- set ns.yaml_data_type = ns.yaml_data_type.split('(', 1)[1].split(',')[1].strip()[:-1] -%}
        {%- break -%}
      {%- endif -%}
    {%- endfor -%}

    {%- if ns.yaml_data_type.startswith('LowCardinality') -%}
      {#-- If contract is LowCardinality, extract the inner type for comparison --#}
      {%- set ns.yaml_data_type = ns.yaml_data_type[15:-1] -%}
    {%- endif -%}

    {%- if sql_data_type != ns.yaml_data_type -%}
      {#-- Column data types don't match #}
      {%- do exceptions.raise_contract_error(yaml_columns, sql_columns) -%}
    {%- endif -%}
```

tests/integration/adapter/constraints/test_constraints.py
```diff
@@ -203,3 +208,49 @@ def models(self):
...
+class TestSpecialTypesConstraints:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "constraints_schema.yml": special_types_schema_yml,
+            "my_model.sql": special_types_schema_sql,
+        }
+
+    def test_special_type_model_constraints_ddl(self, project):
+        run_dbt(["run", "-s", "my_model"])
```
koletzilla, line 257: Can you also add:
- in `test_special_type_model_constraints_ddl ` the valiadtion of `assert country_code_row[1] == 'SimpleAggregateFunction(any, String)'`
- Another test to validate that the `SimpleAggregateFunction` validation may fail if both types are different.

Something like this + adding the missing yaml/sql files would be enough:
```python
class TestSpecialTypesConstraints:
    @pytest.fixture(scope="class")
    def models(self):
        return {
            "constraints_schema.yml": special_types_schema_yml,
            "my_model.sql": special_types_schema_sql,
        }

    def test_special_type_model_constraints_ddl(self, project):
        run_dbt(["run", "-s", "my_model"])

        describe_result = project.run_sql("DESCRIBE TABLE my_model", fetch="all")
        country_code_row = next((row for row in describe_result if row[0] == 'country_code'), None)

        assert country_code_row is not None
        assert country_code_row[1] == 'SimpleAggregateFunction(any, String)'


class TestIncorrectSpecialTypeConstraints:
    @pytest.fixture(scope="class")
    def models(self):
        return {
            "constraints_schema.yml": special_types_schema_yml,
            "my_model.sql": special_types_schema_sql_incorrect_special_type_schema_sql,
        }

    def test_low_cardinality_contract_fails(self, project):
        _, log_output = run_dbt_and_capture(["run", "-s", "my_model"], expect_pass=False)
        assert 'data type mismatch' in log_output.lower()
```

jorgeparavicini:
Hi @koletzilla, great points, I've updated the implementation accordingly. I also modified the macro so that we have a single location where we need to add new "special types". I added both the LowCardinality and the SimpleAggregateFunction there now.

Issue #529, opened 2025-09-30
dbt-clickhouse functionality may suddenly break as dbt-adapter dependency is automatically updated.


koletzilla:
We don't have a defined version of `dbt-adapters` to use. The code in that library may change, breaking some of our functionality like in https://github.com/ClickHouse/dbt-clickhouse/pull/486 where the unit-test functionality was broken. 

This is impacting:
- Our clients: Once they install dbt-clickhouse, they may install an unsupported version of `dbt-adapters`, breaking the functionaility for them.
- I think this has already happened for other features for some clients, like the snapshots.
- This also impact our tests: Github Actions download the latest version of everything we have defined, so tests may break in unrelated PRs.

Solution propose: attach the repo to an specific `dbt-adapters` version and manually update it. Maintained adapters have the version defined in their `pyproject.toml` (our `setup.py`) but they have it without a minor fixed (`"dbt-adapters>=1.7.0,<2.0"`) in our case we should fix the minor to a particular version.

Issue #518, opened 2025-09-17
Case-sensitivity in unit-test in column names failing after upgrading dbt-adapters from 1.16.6 to 1.16.7 (follow-up ticket)


koletzilla:
Looks like there's a bug in the new `dbt-adapters` version that makes some column-names lower-case, breaking the compatibility with ClickHouse as CH uses case-sensitive column names https://github.com/ClickHouse/ClickHouse/issues/33935.

Ticket in the dbt-adapters repository https://github.com/dbt-labs/dbt-adapters/issues/1339

The test that we can use to detect this problem is `dbt.tests.adapter.unit_testing.test_quoted_reserved_word_column_names::BaseUnitTestQuotedReservedWordColumnNames`. We should add it back to the repo once this issue is fixed.

koletzilla:
Clossing this follow up ticket as 1.17 fixes this issue and removes dbt.tests.adapter.unit_testing.test_quoted_reserved_word_column_names::BaseUnitTestQuotedReservedWordColumnNames

PR #486, opened 2025-08-07
Add codec support for column addition in schema changes

Nikita1198:
## Summary

This pull request introduces support for preserving column codec configurations during incremental schema changes in ClickHouse. Previously, when using `on_schema_change: append_new_columns` or `on_schema_change: sync_all_columns`, new columns were added without their specified codec, even if defined in the model's schema configuration. This PR resolves that issue, ensuring codecs are applied as intended.

## Changes Made

- Updated the `clickhouse__add_columns` macro in `schema_changes.sql` to read codec configurations from model column definitions.
- Added logic to include the `CODEC(...)` clause in `ALTER TABLE ADD COLUMN` statements when a codec is specified.
- Ensured backward compatibility: columns without codec specifications function as before.

## Example

```yaml
# schema.yml
models:
  - name: my_model
    columns:
      - name: my_column
        data_type: String
        codec: ZSTD  # This codec is now preserved during schema changes
```

This change ensures that codec configurations, such as `ZSTD`, are retained during incremental schema evolution, addressing the issue of codecs being lost.

## Checklist

- [x] Added unit and integration tests covering common scenarios.
- [x] Provided a human-readable description of the changes for inclusion in the CHANGELOG.
- [ ] Updated documentation in [ClickHouse Docs](https://github.com/ClickHouse/clickhouse-docs) with detailed explanations and tutorials for significant changes.

## Test Coverage

- Added comprehensive integration tests in `test_schema_change_codec.py`.
- Tests cover both `append_new_columns` and `sync_all_columns` scenarios.
- Verified codec presence in the table structure using `SHOW CREATE TABLE`.

PR #530, opened 2025-09-30
Move dbt-adapters as package dependency and lock its version. 

koletzilla:
fix unit-test problem after dbt-adapter was updated.

Closes https://github.com/ClickHouse/dbt-clickhouse/issues/529. <- Check it for context
Closes https://github.com/ClickHouse/dbt-clickhouse/issues/518

Unlocks https://github.com/ClickHouse/dbt-clickhouse/pull/486 as their tests were failing because of this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 #### Bugs
 * Fix Materialized View not dropped when a model's materialization is changed from materialized_view to view ([#516](https://github.com/ClickHouse/dbt-clickhouse/pull/516)).
 * Ensure that temporary tables are not accessed with database clause ([#515](https://github.com/ClickHouse/dbt-clickhouse/pull/515)).
-
+* Lock `dbt-adapters` version to avoid incompatibilities caused by unexpected version upgrades ([#530](https://github.com/ClickHouse/dbt-clickhouse/pull/530)).
 
 ### Release [1.9.3], 2025-09-08
 


dbt/include/clickhouse/macros/materializations/unit.sql
```diff
@@ -4,22 +4,29 @@
...
+  {%- for column_name in tested_expected_column_names -%}
+  {%-   do expected_column_names_quoted.append(column_name_to_quoted[column_name|lower]) -%}
   {%- endfor -%}
 
   {% if not expected_sql %}
-  {%   set expected_sql = get_expected_sql(expected_rows, column_name_to_data_types) %}
+  {%   set expected_sql = get_expected_sql(expected_rows, column_name_to_data_types, column_name_to_quoted) %}
   {% endif %}
-  {% set unit_test_sql = get_unit_test_sql(sql, expected_sql, tested_expected_column_names) %}
+  {% set unit_test_sql = get_unit_test_sql(sql, expected_sql, expected_column_names_quoted) %}
```
koletzilla, line 29: All changes in this file were just copy-pasting the changes they did in their unit.sql file. This is another reason to not to not copy-paste their files :/

setup.py
```diff
@@ -54,6 +54,7 @@ def _dbt_clickhouse_version():
     },
     install_requires=[
         f'dbt-core>={dbt_minor_version}',
+        'dbt-adapters==1.17.0',  # This version should be dbt-adapters>=1.17,<2.0, but keeping it fixed for now to avoid unexpected issues. We need to frequently update it.
```
koletzilla, line 57: We need to even keep the patch version to avoid issues to suddently happen like https://github.com/ClickHouse/dbt-clickhouse/issues/518

koletzilla: Thinking out loud: We can just use `pip lock` to keep all versions fixed and be able to add traceability to the versions upgrades but this feature is still experimental. Lets see if we can do it in the future.

setup.py
```diff
@@ -54,6 +54,7 @@ def _dbt_clickhouse_version():
     },
     install_requires=[
         f'dbt-core>={dbt_minor_version}',
+        'dbt-adapters>=1.17.0,<1.18.0',  # This version should be dbt-adapters>=1.17,<2.0, but keeping it fixed for now to avoid unexpected issues. We need to frequently update it.
```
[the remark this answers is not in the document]
koletzilla: windsurf did not understand the comment

BentsiLeviav (approved):

PR #538, opened 2025-10-03
Start testing with dbt-core 1.10

koletzilla:
## Summary

Related to https://github.com/ClickHouse/dbt-clickhouse/issues/539. You'll find the information and context about what's happening in the ticket.

This PR:
- removes the `dev` dependencies so this don't happen again
- Fixes version 1.10 of dbt-core for dbt-clickhouse 1.9 versions so at least we can make sure 1.11 is not downloaded in the future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ package_name = 'dbt-clickhouse'
 package_version = _dbt_clickhouse_version()
 description = '''The Clickhouse plugin for dbt (data build tool)'''
 
-dbt_minor_version = '1.9'
+dbt_minor_version = '1.10'
 
 if not package_version.startswith(dbt_minor_version):
     raise ValueError(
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -53,7 +53,7 @@ def _dbt_clickhouse_version():
         ]
     },
     install_requires=[
-        f'dbt-core>={dbt_minor_version}',
+        f'dbt-core>={dbt_minor_version},<1.11',
```
koletzilla, line 56: Pending to decide what we are going to do here: move to dbt-core 1.9 (and maybe breaking some users) or just deal with the current status and at least block versions 1.11.*

tests/integration/adapter/concurrency/test_concurrency.py
```diff
@@ -30,4 +30,4 @@ def _validate_results(self, project, results, output):
         check_relations_equal(project.adapter, ["seed", "table_b"])
         check_table_does_not_exist(project.adapter, "invalid")
         check_table_does_not_exist(project.adapter, "skip")
-        assert "PASS=5 WARN=0 ERROR=1 SKIP=1 TOTAL=7" in output
```
koletzilla, line 33: Only test that failed after starting using dbt-core 1.10.* (1.10.13) for testing.

PR #588, opened 2026-01-09
Implement function materialization

shachibista:
## Summary
Adds support for [User Defined Functions](https://docs.getdbt.com/docs/build/udfs).

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Introduces a new relation type and DDL macros for creating ClickHouse functions, plus upgrades to `dbt-core>=1.11.0`; these may affect relation rendering/schema handling and overall adapter compatibility. Connection logging now emits `SQLQuery` events, which could change log volume/behavior.
> 
> **Overview**
> Adds initial support for dbt User Defined Functions on ClickHouse by introducing a `function` relation type (treated as *global* by ignoring schema) and new macros to `CREATE OR REPLACE FUNCTION` for scalar UDFs.
> 
> Upgrades the adapter version to `1.11.0` and bumps development dependency to `dbt-core>=1.11.0`, adds an integration test (`tests/integration/adapter/udf/test_udf.py`) covering basic UDF creation, and updates connection execution to emit `SQLQuery` events (with optional SQL log abridging) before running statements.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 37472f65192a7557475363ab7e6c564ef296b92b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,7 +124,7 @@ For example:
   -- test_ttl.sql
   {{ config(order_by='(ts)', engine='MergeTree()', materialized='table') }}
 
-  SELECT now() AS ts,
+  SELECT now() AS ts, 
         'Some value that should expire!' AS col_ttl
   ```
 
@@ -160,7 +160,7 @@ For example:
 #### Improvements
 * It is now possible to configure a TLS client certificate using `client_cert` and `client_cert_key` profile parameters. ([#413](https://github.com/ClickHouse/dbt-clickhouse/pull/413))
 * Added Support of insert_overwrite in cluster setup with incremental and distributed_incremental materializations ([#394](https://github.com/ClickHouse/dbt-clickhouse/pull/394))
-* Improve index and projections creation process ([#421](https://github.com/ClickHouse/dbt-clickhouse/pull/421))
+* Improve index and projections creation process ([#421](https://github.com/ClickHouse/dbt-clickhouse/pull/421)) 
 
 #### Bugs
 * Reverted breaking changes in MV materialization ([#416](https://github.com/ClickHouse/dbt-clickhouse/pull/416))
@@ -179,7 +179,7 @@ For example:
 
 #### New Features
 * [ClickHouse indexes](https://clickhouse.com/docs/en/optimize/sparse-primary-indexes) are now fully supported for `table` materialization.
-The index config should be added to the model config. for instance:
+The index config should be added to the model config. for instance: 
   ```python
   {{ config(
          materialized='%s',
@@ -203,7 +203,7 @@ The index config should be added to the model config. for instance:
 * Removed support in python 3.8 as it is no longer supported by dbt ([#402](https://github.com/ClickHouse/dbt-clickhouse/pull/402)
 
 ### Bug Fixes
-* Fix a minor bug related to validating existence of an old hanging mv ([#396]())
+* Fix a minor bug related to validating existence of an old hanging mv ([#396]()) 
 
 ### Release [1.8.6], 2024-12-05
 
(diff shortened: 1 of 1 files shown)


koletzilla:
Related to https://github.com/ClickHouse/dbt-clickhouse/issues/590

koletzilla:
Hi @shachibista! Thanks for the contribution. This is a really interesting functionality and I just started checking the implementation details, so I still don't have a full picture of the work needed. I'm sorry I can't give complete feedback yet, but I wanted to ask a few questions to get your opinion:
- Can you walk through all the changes you needed to make to the `materialization function` macro? It looks like we should be able to use practically the same implementation as the default in dbt. Why did you need to apply these changes?
- Were you able to later call this function from a model? I'm getting different errors when I try to use it. Can you add a test to assert that the function can be called from a model?
- It looks like we may need to update dbt-adapters to the latest version to be able to use UDFs, but you didn't do it. Is there a reason not to do it?
- dbt usually releases tests in the dbt-tests-adapter package that we can call from this adapter like this: https://github.com/dbt-labs/dbt-adapters/tree/main/dbt-tests-adapter/src/dbt/tests/adapter/functions. Have you checked if we can reuse these tests?

shachibista:
> It looks like we may need to update dbt-adapters to the latest version to be able to use UDFs, but you didn't do it. Is there a reason not to do it?

Ah yes, I missed that. 

> Can you walk through all the changes you needed to make to the materialization function macro? It looks like we should be able to use practically the same implementation as the default in dbt. Why did you need to apply these changes?

You are correct, I tried to use the default implementation but did not realise that dbt-adapters needed to be updated too. Should be unnecessary after update.

> dbt usually releases tests in the dbt-tests-adapter package that we can call from this adapter like this: https://github.com/dbt-labs/dbt-adapters/tree/main/dbt-tests-adapter/src/dbt/tests/adapter/functions. Have you checked if we can reuse these tests?

Unfortunately we cannot use them because ClickHouse functions expect an expression:

```
CREATE [OR REPLACE] FUNCTION name [ON CLUSTER cluster] AS (parameter0, ...) -> expression
```

However these tests define the function body as a `SELECT` statement.

> Were you able to later call this function from a model? I'm getting different errors when I try to use it. Can you add a test to assert that the function can be called from a model?

What errors are you getting? I did not try to run it from a model but I currently get the following when trying to run from a model.

```
 'SourceQuotingBaseConfig' object has no attribute 'credentials'
```

I am not sure what the reason is.

shachibista:
> About your changes in the extra spaces in the changelog: I'm ok if you want to move the small changelog fixes to a different PR. That way we can reduce the size of this one and we can merge the other asap.

Ah, it was a side-effect of the linter I was using.

> You should be able to override the definition of the files to make them match the CH expectations. That way we should reuse the rest of the tests. Would you try? Maybe we something like this https://github.com/ClickHouse/dbt-clickhouse/blob/main/tests/integration/adapter/unit_testing/test_types.py so the tests functions are reused and we just override the part that are different in CH.

I have now changed the tests to re-use the dbt-adapter tests. Although had to add events so that the tests worked as expected, probably a good thing if other tests depend on it and we want to reuse tests in the future.

> Also another question: The default UDF materialization in dbt mentions it supports both Python and SQL functions. Did you tested both? Can you also add a test to assert that Python one works?

ClickHouse seems to support python UDFs but they [need to be configured](https://clickhouse.com/docs/sql-reference/functions/udf#udf-python) on the server. So I deem it outside dbt's responsibility. I have not tested, nor intend to implement it.

PR #614, opened 2026-02-18
Use trusted publisher to publish packages

### LINKED ISSUE #592: Improving Pypi validations by using a Trusted Publisher
Idea proposed by @mariocj89, documentation on Pypi side https://docs.pypi.org/trusted-publishers/

## What is it
Pypi offers a way to publish to our Pypi account by using a "trusted third part service", in this case Github Actions, instead of just using the current token-authetication way. This has some advantages:
> - Usability: with Trusted Publishing, users no longer need to manually create API tokens on PyPI and copy-paste them into their CI provider. The only manual step is configuring the publisher on PyPI.
> - Security: PyPI's normal API tokens are long-lived, meaning that an attacker who compromises a package's release token can use it until its legitimate user notices and manually revokes it. Trusted Publishing avoids this problem because the tokens minted expire automatically.

This should help with the amount of `Verified details` that Pypi can check in our project. Right now it only verifies the number of Maintainers, but for example it doesn't validates the project links. To view an example of this, you can compare both `Verified details` section in the left column for [dbt-clickhouse](https://pypi.org/project/dbt-clickhouse/) and another repository like [dbt-adapters](https://pypi.org/project/dbt-adapters/)

koletzilla:
## Summary
Closes #592 

With these changes, next time we release a version it will be done using the Trusted Publisher workflow.

We have already configured it in PyPI.

---

> [!NOTE]
> **Low Risk**
> Release automation changes only; main risk is misconfigured GitHub/PyPI OIDC permissions causing publish failures, not runtime/user-facing behavior.
> 
> **Overview**
> Switches the PyPI release GitHub Actions workflow to **PyPI Trusted Publishing** by granting `id-token: write` and removing the token/password secret-based publish configuration.
> 
> Updates metadata/docs to reflect the new release process: adds a `1.10.1` changelog entry about trusted publishing and expands `pyproject.toml` `project.urls` with documentation, changelog, and community links.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit eea43a27029cf46705f52f1bc0fb0858d62b25bc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ Repository = "https://github.com/ClickHouse/dbt-clickhouse"
 Issues = "https://github.com/ClickHouse/dbt-clickhouse/issues"
 Documentation = "https://github.com/ClickHouse/dbt-clickhouse#documentation"
 Changelog = "https://github.com/ClickHouse/dbt-clickhouse/blob/main/CHANGELOG.md"
-slack = "https://clickhouse.com/slack"
+Slack = "https://clickhouse.com/slack"
 
 [tool.setuptools]
 packages = {find = {include = ["dbt", "dbt.*"], namespaces = true}}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,9 @@ dependencies = [
 Homepage = "https://github.com/ClickHouse/dbt-clickhouse"
 Repository = "https://github.com/ClickHouse/dbt-clickhouse"
 Issues = "https://github.com/ClickHouse/dbt-clickhouse/issues"
+Documentation = "https://github.com/ClickHouse/dbt-clickhouse#documentation"
+Changelog = "https://github.com/ClickHouse/dbt-clickhouse/blob/main/CHANGELOG.md"
+slack = "https://clickhouse.com/slack"
 
 [tool.setuptools]
 packages = {find = {include = ["dbt", "dbt.*"], namespaces = true}}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ Homepage = "https://github.com/ClickHouse/dbt-clickhouse"
 Repository = "https://github.com/ClickHouse/dbt-clickhouse"
 Issues = "https://github.com/ClickHouse/dbt-clickhouse/issues"
 Documentation = "https://github.com/ClickHouse/dbt-clickhouse#documentation"
+"ClickHouse Docs" = "https://clickhouse.com/docs/integrations/dbt"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -33,6 +33,9 @@ dependencies = [
 Homepage = "https://github.com/ClickHouse/dbt-clickhouse"
 Repository = "https://github.com/ClickHouse/dbt-clickhouse"
 Issues = "https://github.com/ClickHouse/dbt-clickhouse/issues"
+Documentation = "https://github.com/ClickHouse/dbt-clickhouse#documentation"
+Changelog = "https://github.com/ClickHouse/dbt-clickhouse/blob/main/CHANGELOG.md"
+slack = "https://clickhouse.com/slack"
```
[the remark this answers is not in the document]
koletzilla: Not really needed as this part is case insensible, but fixing to add consistency

pyproject.toml
```diff
@@ -33,6 +33,9 @@ dependencies = [
 Homepage = "https://github.com/ClickHouse/dbt-clickhouse"
 Repository = "https://github.com/ClickHouse/dbt-clickhouse"
 Issues = "https://github.com/ClickHouse/dbt-clickhouse/issues"
+Documentation = "https://github.com/ClickHouse/dbt-clickhouse#documentation"
```
BentsiLeviav, line 36: Can we add the ClickHouse.com documentation page here? Or github is mandatory?

koletzilla: There's an issue on how many urls can PyPI verify. [PyPI only verifies URLs coming from other Pypi urls or from Github](https://docs.pypi.org/project_metadata/#:~:text=JSON%20API.-,Verified%20details,-PyPI%20currently%20supports). If we add a clickhouse.com url, it will not consider that url as verified. This is the image from test.pypi.org where we can see which things get verified. This image was take when the documentation still pointed to clickhouse.com:

Screenshot 2026-02-20 at 11 39 47

I don't mind seeing the slack as unverified, but it didn't looked right to have the documentation unverified. This workaround is a bit ugly, but it let us have this link with the rest of the "verified" ones

BentsiLeviav: Can we have 2 documentation links? One would be named documentation (pointing to the readme file), and another would be an unferfied link named ClickHouse Docs linking to ClickHouse.com? 
If not, LGTM

BentsiLeviav (approved):

pyproject.toml
koletzilla: Added!

Issue #544, opened 2025-10-12
1.9.3 broke snapshots with compound unique keys


agrueneberg:
The new snapshot code introduced in #509 does not account for compound unique keys, introducing string arrays (e.g., `['date', 'team_id', 'person_id']`) as `dbt_unique_key` in both `snapshotted_data` and `insertions_source_data`. When combined with the `LEFT JOIN` used in `insertions`, this leads to an unbounded Cartesian product — resulting in infinite insertions until disk space is exhausted.

Issue #291, opened 2024-06-09
snapshot invalidate_hard_deletes feature


dstsimokha:
**Is your feature request related to a problem? Please describe.**
invalidate_hard_deletes=True for snapshots not working with CH because of its default non-nullable column

**Describe the solution you'd like**
dbt general `default__snapshot_staging_table` macro can be modified in two places by changing `{{ strategy.unique_key }}` to `toNullable({{ strategy.unique_key }})` - in `snapshotted_data` and `deletes_source_data` CTEs

rjoelnorgren:
I was able to workaround this by specifying `query_settings={"join_use_nulls": 1}` in the snapshot config.

rjoelnorgren:
Encountered an issue with the suggested work around after upgrading from 1.8.3 -> 1.8.7. The query_settings are now included in the tmp snapshot creation ddl. As a result, it throws: 
```
DB::Exception: Setting allow_nullable_key is neither a builtin setting nor started with the prefix 'SQL_' registered for user-defined settings.
```

It is now also necessary to set `settings: {"allow_nullable_key": 1}`

Issue #271, opened 2024-04-18
invalidate_hard_deletes option does not work with clickhouse


andrewkomkov:
### Describe the bug
invalidate_hard_deletes option (https://docs.getdbt.com/reference/resource-configs/invalidate_hard_deletes) does not work with clickhouse
### Steps to reproduce
1. Create snapshot like this
```
{% snapshot snap_imdb_actors %}
    {{
        config(
          target_schema='snapshots',
          strategy='check',
          unique_key='id',
          check_cols='all',
          invalidate_hard_deletes=True
        )
    }}
    with actors as (
    select * from {{ source('imdb', 'actors') }})

    select * from actors
{% endsnapshot %}
```
2. Run it
3. Then delete a row from table imdb.actors
4. run snapshot again

### Expected behaviour
In snapshot table, row with deleted id should have dbt_valid_to, but it does null

#### Environment
* dbt version: 1.7.12
* dbt-clickhouse version: 1.7.6

andrewkomkov:
I guess i found cause of the problem
If somehow pass this setting SET join_use_nulls = 1;
It will wokr
The question is - how can i pass settings like this to dbt?

andrewkomkov:
![image](https://github.com/ClickHouse/dbt-clickhouse/assets/33517893/78596699-f71d-49a4-a347-32884e0ae1d3)

Fixed with this setting

genzgd:
thanks for the investigation -- my sense is that dbt-clickhouse should make "join_use_nulls" the default perhaps?

andrewkomkov:
I don't think you should make it enabled by default.
Since it is not enabled by default in Clickhouse itself
However, this should be included in the module description

PR #630, opened 2026-04-02
fix(snapshot): support dbt_valid_to_current config in snapshot macros

### LINKED ISSUE #481: dbt snapshot - check strategy - stopped working properly
Snapshot check strategy odd behavior. While timestamp strategy works fine, snapshot check strategy works quite odd and works fine only if _dbt_updated_at_ parameter in .yml config is used (similar to timestamp strategy).
Main issue is with inconsistent results - historic version rows are not closed (they remain valid to 2100-01-01), new version rows are automatically closed (they are not valid to 2100-01-01), etc.

### Steps to reproduce
1. Create table
2. Run initial snapshot - everything works
3. Update some column, for example "company_name"
4. Run snapshot again - as stated above - old version rows are not closed, new version rows are automatically closed, etc.

### Expected behavior
dbt_valid_to = some date/time - ie. actual time of snapshot run - historic version
dbt_valid_to = 2100-01-01 00:00:00 for opened current versions

#### Environment
* dbt version: 1.9.4
* dbt-clickhouse version: 1.9.1

#### ClickHouse server
* ClickHouse Server version: 25.6.3.116

### Snapshot model
```
{% snapshot snp_test_table %}

{{ config(
    query_settings = {'join_use_nulls': 1},
    target_schema=generate_schema_name('snapshots')
    )
}}

SELECT
    {{ dbt_utils.star(
      from = ref('stg_test_table')
    ) }}
FROM {{ ref('stg_test_table') }}

{% endsnapshot %}
```

### Snapshot yml
```
version: 2
snapshots:
  - name: snp_test_table
    description: 'Test table'
    latest_version: 1
    versions:
      - v: 1
        config:
          alias: snp_test_table
    config:
      unique_key: vat_number
      strategy: check
      dbt_valid_to_current: "{{ var('future_proof_date') }}"
      hard_deletes: 'invalidate'
      check_cols:
        - company_name
        - post_code
        - company_city
      group: TGDA
      tags:
        - test_table
```

var('future_proof_date') is defined in the dbt_project.yml file as:
`vars:
  future_proof_date: "toDateTime('2100-01-01 00:00:00')"`

#### DDL's

```
CREATE TABLE services.test_table
(

    `vat_number` String,

    `company_name` String,

    `post_code` String,

    `company_city` String
)
ENGINE = MergeTree
ORDER BY vat_number
SETTINGS index_granularity = 8192;
```

PSalinovic:
## Summary
The ClickHouse adapter's custom snapshot staging table macros for both
timestamp and check strategies ignored the dbt_valid_to_current config
option. Current records were always filtered with WHERE dbt_valid_to IS
NULL, which missed rows that had the configured sentinel value (e.g.
'9999-12-31'). This caused duplicate rows on every subsequent snapshot
run when dbt_valid_to_current was configured.

The fix reads config.get('dbt_valid_to_current') and:
- Filters current records with (dbt_valid_to = <value> OR dbt_valid_to IS NULL)
- Sets dbt_valid_to for new insertions using coalesce(nullif(x, x), <value>)
The OR IS NULL fallback ensures backward compatibility with rows
snapshotted before dbt_valid_to_current was configured.

Fixes: #481

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 #### Improvements
 * Starting with this release the `dbt-clickhouse` packages will be published to PyPI using Github Actions as a [Trusted Publisher](https://docs.pypi.org/trusted-publishers/). This will improve both the usability and the security of the release process ([#614](https://github.com/ClickHouse/dbt-clickhouse/pull/614)).
 
+#### Bugs
+* Fix `dbt_valid_to_current` snapshot configuration being ignored in the ClickHouse adapter. The snapshot macros for both timestamp and check strategies now correctly read and apply the `dbt_valid_to_current` config value, matching dbt-core's expected behavior. Previously, snapshots configured with `dbt_valid_to_current` would produce duplicate rows on subsequent runs because the adapter always filtered current records with `WHERE dbt_valid_to IS NULL`, missing rows that had the configured sentinel value.
+
 
 ### Release [1.10.0], 2026-02-16
 
--- a/dbt/include/clickhouse/macros/materializations/snapshot.sql
+++ b/dbt/include/clickhouse/macros/materializations/snapshot.sql
@@ -99,6 +99,8 @@
 
 {% macro clickhouse__snapshot_staging_table_check_strategy(strategy, source_sql, target_relation) -%}
 
+    {% set dbt_valid_to_current = config.get('dbt_valid_to_current') %}
+
     with snapshot_time as (
         select {{ strategy.updated_at }} as ts  -- Single timestamp
     ),
@@ -114,7 +116,12 @@
             {{ strategy.unique_key }} as dbt_unique_key
 
         from {{ target_relation }}
-        where dbt_valid_to is null
+        where
+            {% if dbt_valid_to_current %}
+                ( dbt_valid_to = {{ dbt_valid_to_current }} or dbt_valid_to is null )
+            {% else %}
+                dbt_valid_to is null
+            {% endif %}
 
     ),
 
@@ -125,7 +132,7 @@
             {{ strategy.unique_key }} as dbt_unique_key,
             snapshot_time.ts as dbt_updated_at,
             snapshot_time.ts as dbt_valid_from,
-            nullif(snapshot_time.ts, snapshot_time.ts) as dbt_valid_to,
+            coalesce(nullif(snapshot_time.ts, snapshot_time.ts), {{ dbt_valid_to_current or 'null' }}) as dbt_valid_to,
             {{ strategy.scd_id }} as dbt_scd_id
 
         from snapshot_query, snapshot_time
@@ -217,6 +224,8 @@
 
 {% macro clickhouse__snapshot_staging_table_timestamp_strategy(strategy, source_sql, target_relation) -%}
 
+    {% set dbt_valid_to_current = config.get('dbt_valid_to_current') %}
+
     with snapshot_query as (
 
         {{ source_sql }}
@@ -229,7 +238,12 @@
             {{ strategy.unique_key }} as dbt_unique_key
 
         from {{ target_relation }}
-        where dbt_valid_to is null
+        where
+            {% if dbt_valid_to_current %}
+                ( dbt_valid_to = {{ dbt_valid_to_current }} or dbt_valid_to is null )
+            {% else %}
+                dbt_valid_to is null
+            {% endif %}
 
     ),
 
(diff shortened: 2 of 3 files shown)


koletzilla (changes requested):
## Review Summary

The fix correctly addresses the `dbt_valid_to_current` gap in ClickHouse's custom snapshot staging macros. The approach aligns with dbt-core's `default__snapshot_staging_table` pattern and the merge macro correctly doesn't need changes (it joins on `dbt_scd_id` which is unique per row, unlike dbt-core's `MERGE` which filters on `dbt_valid_to`).

### What looks good

- **WHERE clause** matches dbt-core: `(dbt_valid_to = <sentinel> OR dbt_valid_to IS NULL)` — the IS NULL fallback ensures backward compatibility with rows snapshotted before the config was added.
- **New insert `dbt_valid_to`**: `coalesce(nullif(ts, ts), <sentinel_or_null>)` matches dbt-core's `get_dbt_valid_to_current` macro.
- **Updates correctly expire with the real timestamp**, not the sentinel — this is correct SCD2 behavior.
- **First-run path** is unaffected — dbt-core's `default__build_snapshot_table` (which already handles `dbt_valid_to_current`) is used since the adapter doesn't override it.
- Both check and timestamp strategies are handled.

### Issues

#### 1. Missing test for record modifications (high)

This is the **exact scenario from issue #481**: "Update some column → old version rows are not closed, new version rows are automatically closed." The tests only cover inserts and idempotent re-runs — they never modify an existing row and re-snapshot.

Please add a test step that:
1. Modifies an existing row's value (e.g., update a seed CSV with a changed name for id=1)
2. Re-runs snapshot
3. Asserts the **old** version has `dbt_valid_to` = real timestamp (not sentinel, not NULL)
4. Asserts the **new** version has `dbt_valid_to` = sentinel
5. Asserts total row count = N + 1

Without this, the most important SCD2 behavior is unverified.

#### 2. `hard_deletes: 'invalidate'` not tested with `dbt_valid_to_current` (medium)

Issue #481 specifically configures `hard_deletes: 'invalidate'`. The fix correctly filters current records in the `snapshotted_data` CTE (which the `deletes` CTE depends on), but there's no test covering this combination.

#### 3. Minor: `equals()` macro not used (low, informational)

dbt-core uses `{{ equals(column, value) }}` for the sentinel comparison (which provides NULL-safe equality depending on behavior flags). This PR uses direct `=`. In ClickHouse this works correctly (`NULL = X` evaluates to NULL/falsy, caught by the `OR IS NULL` fallback), but noting the divergence.

koletzilla:
Hi @PSalinovic ! Thanks for the contribution, those snapshot issues are quite relevant, so having someone working on them is more than welcome! 

We are using AI to help us reviewing PRs as a first check before going deep intro a manual validation. We don't blindly trust them, but usually they find some stuff interesting and in this case it looks like it has some opinions on changes that look good. Can you check them?

About the CI:
- There're also a couple of small lint problems, easy to fix.
- test_cloud fails in all community PRs, so ignore it. It's not needed to merge this PR.

PSalinovic:
Hi, thanks for the feedback.

I fixed the linting issue from the failing test.

Also, I added two more test cases in the test file. That was a good catch by the clanker. Testing the case with the updated record works fine and the test passes however the test with the delete that uses 'hard_deletes' config variable does not pass. It looks like this is a known bug that already has two issues opened: #271 and #291. The test will actually pass if I have the join_use_nulls set to true in the global config, which i did locally by updating the conftest.py. I didn't have the time to look into this more but it seems like it could be solved by using NOT EXIST instead of LEFT JOIN. I'll try to see if this works and submit a PR if it does.
I left the test in the test file but disabled it and left a comment.

Another issue I encountered is that specifying composite keys in the snapshot config per dbt docs does not work and acutally produces a CROSS JOIN where the data multiplies. This is mentioned in #630 and #544. I'll try to submit a PR for this as well as this is critical on our end.

koletzilla:
Hi @PSalinovic , sorry for the lack of updates on your PR.

I'm getting back to it and while I was reviewing it I realised we may need to make more changes in that codebase that may invalidate your current PR. I think we may be able to merge your PR, but as many changes are coming, I think I prefer to really understand all the needed changes for Snapshots first.

We talked internally and we want to dedicate some time on "revamping" the Snapshots feature, including missing functions and fixing as many Snapshot issues as the ones in the huge list https://github.com/ClickHouse/dbt-clickhouse/issues?q=is%3Aissue%20state%3Aopen%20snapshot

I think we may release a new dbt-clickhouse version with the current released stuff + maybe some of the other minor fixes, but I'll merge this PR & work on other snapshots one in the next one.

I'll keep you posted.

PR #632, opened 2026-04-15
feat: populate query_id in AdapterResponse from ClickHouse driver

### LINKED ISSUE #617: Populate ClickHouse query_id in AdapterResponse
### Description

The dbt-core `AdapterResponse` dataclass already has a `query_id` field:

```python
@dataclass
class AdapterResponse(dbtClassMixin):
    _message: str
    code: Optional[str] = None
    rows_affected: Optional[int] = None
    query_id: Optional[str] = None
```

However, the dbt-clickhouse adapter's `execute()` method in `connections.py` never sets it:

```python
return AdapterResponse(_message=status), table
```

The underlying `clickhouse-connect` driver already captures the ClickHouse `query_id` from the `X-ClickHouse-Query-Id` response header and exposes it via `QueryResult.query_id` and `QuerySummary.summary['query_id']`. It's just never passed through to the `AdapterResponse`.

### Motivation

The `query_id` is essential for correlating dbt model executions with ClickHouse's `system.query_log`. This enables:

- Debugging slow or failed model runs by looking up the exact query in ClickHouse's query log
- Tracking resource consumption (memory, read bytes, CPU time) per dbt model
- Linking dbt run results to ClickHouse query profiling data

Without `query_id` in the adapter response, users have to resort to fragile timestamp-based correlation with `system.query_log`, which is unreliable when multiple queries run concurrently.

### Proposed Change

In `dbt/adapters/clickhouse/connections.py`, the `execute()` method should extract `query_id` from the driver result and pass it to `AdapterResponse`:

```python
def execute(self, sql, auto_begin=False, fetch=False, limit=None):
    ...
    with self.exception_handler(sql):
        ...
        if fetch:
            query_result = client.query(sql)
        else:
            query_result = client.command(sql)
        status = self.get_status(client)
        ...

        query_id = ''
        if fetch and hasattr(query_result, 'query_id'):
            query_id = query_result.query_id or ''
        elif isinstance(query_result, QuerySummary):
            query_id = query_result.summary.get('query_id', '')

        return AdapterResponse(_message=status, query_id=query_id), table
```

### Environment

- **dbt-clickhouse version**: 1.9.5
- **clickhouse-connect version**: (bundled with dbt-clickhouse)
- **dbt-core version**: 1.9.x

I also checked in the latest version of dbt-clickhouse, and didn't find a way to get the query_id.

shubham1166:
## Summary

Closes #617

The `execute()` method in `connections.py` now extracts `query_id` from the ClickHouse driver result and passes it through to `AdapterResponse`.

### How it works

The dbt-core `AdapterResponse` dataclass already has a `query_id` field but it was never populated. The fix uses `hasattr` checks to handle both client types gracefully:

- **HTTP client (clickhouse-connect), `fetch=True`**: reads `QueryResult.query_id`
- **HTTP client (clickhouse-connect), `fetch=False`**: reads `QuerySummary.summary['query_id']`
- **Native client (clickhouse-driver)**: `query_id` is not exposed by the driver — gracefully falls back to empty string

No new imports are required. The checks are non-breaking for both client types.

### Changes

- `dbt/adapters/clickhouse/connections.py`: extract and pass `query_id` in `execute()`
- `CHANGELOG.md`: add entry under upcoming `1.10.1` release

### Why this matters

Without `query_id` in the adapter response, users must rely on fragile timestamp-based correlation with `system.query_log`. With this change, dbt model executions can be directly correlated with ClickHouse's query log for debugging slow runs, tracking memory/CPU usage per model, and linking dbt results to ClickHouse query profiling data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### Release [1.10.1], 2026-0X-XX
 
 #### Improvements
+* Populate `query_id` in `AdapterResponse` from the ClickHouse driver result. The HTTP client exposes `query_id` via `QueryResult.query_id` (fetch queries) and `QuerySummary.summary['query_id']` (command queries). This enables users to correlate dbt model executions with `system.query_log` for debugging, resource tracking, and query profiling ([#617](https://github.com/ClickHouse/dbt-clickhouse/issues/617)).
 * Starting with this release the `dbt-clickhouse` packages will be published to PyPI using Github Actions as a [Trusted Publisher](https://docs.pypi.org/trusted-publishers/). This will improve both the usability and the security of the release process ([#614](https://github.com/ClickHouse/dbt-clickhouse/pull/614)).
 
 
--- a/dbt/adapters/clickhouse/connections.py
+++ b/dbt/adapters/clickhouse/connections.py
@@ -103,7 +103,16 @@ class ClickHouseConnectionManager(SQLConnectionManager):
                 from dbt_common.clients.agate_helper import empty_table
 
                 table = empty_table()
-            return AdapterResponse(_message=status), table
+
+            query_id = ''
+            if fetch and hasattr(query_result, 'query_id'):
+                query_id = query_result.query_id or ''
+            elif not fetch and hasattr(query_result, 'summary') and isinstance(
+                query_result.summary, dict
+            ):
+                query_id = query_result.summary.get('query_id', '')
+
+            return AdapterResponse(_message=status, query_id=query_id), table
 
     def add_query(
         self,
--- a/dbt/adapters/clickhouse/connections.py
+++ b/dbt/adapters/clickhouse/connections.py
@@ -107,8 +107,10 @@ class ClickHouseConnectionManager(SQLConnectionManager):
             query_id = ''
             if fetch and hasattr(query_result, 'query_id'):
                 query_id = query_result.query_id or ''
-            elif not fetch and hasattr(query_result, 'summary') and isinstance(
-                query_result.summary, dict
+            elif (
+                not fetch
+                and hasattr(query_result, 'summary')
+                and isinstance(query_result.summary, dict)
             ):
                 query_id = query_result.summary.get('query_id', '')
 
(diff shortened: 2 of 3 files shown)


shubham1166:
Hey, i will fix the tests and lints(will add new tests also). Will replicate it locally. If there is anything i think i have skipped, please let me know.

koletzilla:
Hi @shubham1166! Don't worry about cloud testing, those don't work on fork's PRs (we need to fix this). The important ones in this case are the others and they look great 👍 

The test suggested by copilot looks interesting, feel free try to implement it. Don't blindly follow Copilot review, if you find the test is not useful or you find another way to test it, write it in a comment.

After you finish doing these changes, I'll properly review it.

Thanks!

shubham1166:
Thanks! I've added unit tests covering all four paths — HTTP fetch, HTTP command, and the native client fallback for both. They're mocked unit tests rather than integration tests since the existing test suite follows that pattern. I'm aware these are slightly circular since we're mocking the inputs ourselves, so they document intended behaviour rather than proving end-to-end correctness. I want to hear if you have any better recommendation. For me i am just trying to learn a bit more.

koletzilla:
Hi @shubham1166! Thanks again for the work on this, and especially for taking the time to add tests, that's always appreciated.

I've been reviewing your PR carefully. The implementation looks good for the HTTP path, but I got stuck trying to extend it to the Native connection. As you noted, `query_id is not exposed by the driver`, and after digging in I couldn't find a clean way to surface it. The response the native client returns is pretty tightly coupled with the final data, so threading the query_id back through wasn't straightforward.

While I was investigating, #634 came in with a different angle: instead of relying on a query_id returned by ClickHouse, generate one on the dbt side and send it with the query. That ends up fixing the issue for both connectors in one go, which is why we're going to move forward with that PR.

I want to be clear that your PR here genuinely helped, surfacing the native-driver limitation is part of what made it obvious that a different approach was needed. I really appreciate the contribution, and I hope you'll come back and pick up another issue. Thanks again!

PR #640, opened 2026-05-07
[POC] Fix documentation generation by internally filling the database configuration field

### LINKED ISSUE #311: Database tab is empty in generated docs
### Describe the bug
When serving the generated docs, I only have models under the Project tab. The Database tab is completely empty. Is this due to that dbt expects database.schema.table while ClickHouse skips the schema part?

Is this a concious choice or is my dbt version misconfigured?

### Steps to reproduce
1. run `dbt docs generate`
2. run `dbt docs serve`
3. Browse to the tab Overview -> Database

### Expected behaviour
I was expecting to look at trees of databases and tables.

#### Environment
* dbt version: 1.7.12
* dbt-clickhouse version: 1.7.6
* clickhouse-connect version (if using http): 0.7.0
* Python version: 3.11.9
* Operating system: macOS-14.5-arm64-arm-64bit

#### ClickHouse server
* ClickHouse Server version: 24.5.1.1763

koletzilla:
# DO NOT MERGE, IT'S NOT SAFE.

This is just a POC to try to understand how many changes would be required to fix https://github.com/ClickHouse/dbt-clickhouse/issues/311

The issue is related to our dbt-clickhouse not filling the `database` field on purpose. dbt uses database, schema and table to reference resources crates in the database but CH only uses 2, database and table. in dbt world, we are using instead schema and table and leave empty database to avoid errors.

This impacts tools that may want to process this information to build information systems.

In this PR I'm trying to see how we can fill the database configuration in some places so:
- 🟡 Users will still not configure it on purpose, only schema should be available and used
- 🟡 Code internally should NOT use the database field, only the schema is the one that needs to be used.
- ✅ The database (being the same name as the schema) is filled in catalog.json and manifest.json files son the `database` tab in the docs actually have data.

We need to make sure all these things are safe before moving forward with this PR

### DIFF
--- a/dbt/adapters/clickhouse/credentials.py
+++ b/dbt/adapters/clickhouse/credentials.py
@@ -56,7 +56,13 @@ class ClickHouseCredentials(Credentials):
                 f'On Clickhouse, database must be omitted or have the same value as'
                 f' schema.'
             )
-        self.database = ''
+        # ClickHouse has a single namespace ("database") which the adapter maps
+        # to dbt's "schema". Mirror the schema onto `database` so that
+        # `target.database` and source nodes (whose `database` is sourced from
+        # credentials when not explicitly set) stay in sync with the actual
+        # ClickHouse database. This keeps `manifest.json`, `catalog.json` and
+        # `dbt docs generate` consistent with the data warehouse.
+        self.database = self.schema or ''
 
         # clickhouse_driver expects tcp_keepalive to be a tuple if it's not a boolean
         if isinstance(self.tcp_keepalive, list):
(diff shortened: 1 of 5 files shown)


koletzilla:
Note: as expected with this approach as mentioned here https://github.com/ClickHouse/dbt-clickhouse/issues/311#issuecomment-2955622423 , now some relations are wrongly named in the docs by including both schema and database parts like `default.default.my_first_dbt_model`

koletzilla:
Closing as this is something we should defintively takling in other ways as one of the mentioned issues is still present: In any case, the docs forces us on having the 3 elements, but CH can only expose 2 or it will be misleading

Issue #375, opened 2024-10-31
When using dbt-expectations with Clickhouse, it can't convert the timezone when tests require a datetime column


redecvi:
### Describe the bug
When using dbt-expectations with Clickhouse, it can't convert the timezone when tests require a datetime column. It seems that the convert_timezone function of dbt-date is not set up / specified for Clickhouse, so it uses the default command which is not compatible with Clickhouse.

### Steps to reproduce
1. Set up dbt with Clickhouse
2. Add catalogica/dbt-expecations with dbt deps
3. Add a test that uses a datetime column, for example:
`  tables:
  - name: test_table
    columns:
      - name: date_col
        tests:
          - dbt_expectations.expect_row_values_to_have_recent_data:
              datepart: day
              interval: 1`
4. Following error will appear:
`Database Error in test dbt_expectations_source_expect_row_values_to_have_recent_data_dev_test_table_date_col__day__1 (models/gold/schema.yml)
  HTTPDriver for https://mc4yb4j21e.europe-west4.gcp.clickhouse.cloud:8443 returned response code 404)
   Code: 46. DB::Exception: Unknown function convert_timezone. Maybe you meant: ['serverTimezone','serverTimeZone']: While processing CAST(date_col, 'timestamp') <= convert_timezone('UTC', 'UTC', CAST(now(), 'DateTime')). (UNKNOWN_FUNCTION) (version 24.6.1.4609 (official build))
`

### Expected behaviour
Test will run and process the timestamp column correctly

### Code examples, such as models or profile settings
schema.yml
`version: 2

sources:
- name: dev
  tables:
  - name: test_table
    columns:
      - name: date_col
        tests:
          - dbt_expectations.expect_row_values_to_have_recent_data:
              datepart: day
              interval: 1`

### dbt and/or ClickHouse server logs
10:09:13  Running with dbt=1.8.7
10:09:13  Registered adapter: clickhouse=1.8.4
10:09:13  Found 1 model, 1 test, 1 source, 725 macros
10:09:13  
10:09:14  Concurrency: 1 threads (target='dev')
10:09:14  
10:09:14  1 of 1 START test dbt_expectations_source_expect_row_values_to_have_recent_data_dev_test_table_date_col__day__1  [RUN]
10:09:14  1 of 1 ERROR dbt_expectations_source_expect_row_values_to_have_recent_data_dev_test_table_date_col__day__1  [ERROR in 0.06s]
10:09:14  
[...]

PR #646, opened 2026-05-14
feat: add clickhouse__convert_timezone for dbt_date compatibility

dataders:
## Summary

The [`dbt_date` package](https://github.com/calogica/dbt-date) provides a `convert_timezone()` macro that dispatches to `{adapter}__convert_timezone`. ClickHouse is not currently supported, causing `{{ dbt_date.convert_timezone(...) }}` calls to fail with an "Unknown macro" or unsupported adapter error.

This PR adds `clickhouse__convert_timezone` to `utils/timestamps.sql` alongside the existing timestamp utilities.

## Implementation

```jinja
{%- macro clickhouse__convert_timezone(column, target_tz, source_tz="UTC") -%}
    toTimeZone(
        toDateTime({{ column }}, '{{ source_tz }}'),
        '{{ target_tz }}'
    )
{%- endmacro -%}
```

Uses ClickHouse's native `toDateTime(col, tz)` to interpret the source timezone, then `toTimeZone(dt, tz)` to convert — the idiomatic ClickHouse approach.

This implementation was already independently discovered in the [`ClickHouse/jaffle-shop-clickhouse`](https://github.com/ClickHouse/jaffle-shop-clickhouse/blob/main/macros/dbt_date_ch.sql) project, where it lives as a project-level workaround. Moving it into the adapter itself means all dbt-clickhouse users get `dbt_date.convert_timezone()` support without needing a manual override.

## Testing

Confirmed working against ClickHouse 24.3 via the [dbt Fusion ClickHouse adapter](https://github.com/dbt-labs/dbt-core/issues/13168) e2e test suite. The `metricflow_time_spine` model (which uses `dbt_date.convert_timezone`) compiles and runs correctly with this macro in place.

### DIFF
--- a/dbt/include/clickhouse/macros/utils/timestamps.sql
+++ b/dbt/include/clickhouse/macros/utils/timestamps.sql
@@ -6,3 +6,10 @@
   {%- set result = "toDateTime('" ~ timestamp ~ "')" -%}
   {{ return(result) }}
 {%- endmacro %}
+
+{%- macro clickhouse__convert_timezone(column, target_tz, source_tz="UTC") -%}
+    toTimeZone(
+        toDateTime({{ column }}, '{{ source_tz }}'),
+        '{{ target_tz }}'
+    )
+{%- endmacro -%}


koletzilla:
Thanks for the contribution 🙇 

I see the `dbt_date` package currently points to the godatadriven repo https://hub.getdbt.com/godatadriven/dbt_date/latest/. Inside it we can find many implementations from this macro for different adapters https://github.com/godatadriven/dbt-date/blob/2e4bddaafa60122ca5d608c13d2225a59ef439fd/macros/calendar_date/convert_timezone.sql#L4. I think instead of adding it here, we should try to add this new macro in that repo and use that package. Would yo move this new macro to that file? Thanks!

> Confirmed working against ClickHouse 24.3 via the https://github.com/dbt-labs/dbt-core/issues/13168 e2e test suite. The metricflow_time_spine model (which uses dbt_date.convert_timezone) compiles and runs correctly with this macro in place.

Not related to this PR but wanted to comment it: we should not test anymore against a version that old. We should test with the supported versions https://github.com/clickhouse/clickhouse/security or if we want to pick just one, use the latest LTS, which currently is 26.3

koletzilla:
Referencing here https://github.com/ClickHouse/dbt-clickhouse/issues/375 as we may be able to officially solve it now.

koletzilla:
Closing this PR in favour of just adding support for ClickHouse to the whole dbt-date module https://github.com/godatadriven/dbt-date/pull/56

Issue #615, opened 2026-02-19
Deprecate `cluster_mode` setting and simplify EXCHANGE checking


koletzilla:
**Is your feature request related to a problem? Please describe.**
`cluster_mode` is not relevant anymore. It applies to these things:
- Set `database_replicated_enforce_synchronous_settings = 1`: Only applies to Replicated database engines
- Set `insert_quorum', 'auto'`: Only applies to Replicated database engines.
- Decide when to check if EXCHANGE is available: It doesn't depend on the `cluster_mode` but only on the database_engine.

EXCHANGE doesn't need to take `cluster_mode` into account: It depends on the database_engine and on the current system. 
We also need to update the 

**Describe the solution you'd like**
- Deprecate `cluster_mode` completely so:
  - Replicated settings only apply to that database engine
  - EXCHANGE settings don't depend on it. Apart from this, we may benefit from setting `check_exchange ` as `False` now as this should be really uncommon (only old Linux versions or directly running ClickHouse in mac without a container). Users will be able to still set it to true in these envs.

We also need to update the docs with the changes we do.

PR #653, opened 2026-05-24
perf: Short-circuit exchange check on Shared DB engine and run whole check only in one thread

### LINKED ISSUE #641: `_check_atomic_exchange` probe leaks orphan tables when cleanup fails (CH Cloud, v1.10.0)
## Summary

The `_check_atomic_exchange` probe in `dbclient.py` leaves `__dbt_exchange_test_<0|1>_<uuid>` tables behind on the failure side — when a DROP in the cleanup loop raises `DbtDatabaseError`, the rest of the loop is skipped and the remaining probe table(s) leak.

The probe also runs **once per connection** (so many times per `dbt run`), so any environment that fails the EXCHANGE accumulates orphans linearly with run count. In one Argent/dbt-clickhouse pipeline against ClickHouse Cloud (v25.12.1.1497) on a daily refresh schedule, this produced **108 orphan `__dbt_exchange_test_*` tables in a single database** over ~2 weeks before we noticed and added a periodic cleanup sweep.

## Environment

- `dbt-clickhouse`: 1.10.0
- `dbt-core`: 1.11.7
- ClickHouse Cloud: 25.12.1.1497 (Shared / SharedMergeTree)
- Python: 3.12

## Where

[`dbt/adapters/clickhouse/dbclient.py::_check_atomic_exchange`](https://github.com/ClickHouse/dbt-clickhouse/blob/v1.10.0/dbt/adapters/clickhouse/dbclient.py#L227-L259)

## What happens

The probe creates two `__dbt_exchange_test_<0|1>_<uuid>` tables, attempts `EXCHANGE TABLES`, and cleans up in a `finally`:

```python
finally:
    try:
        for table in swap_tables:
            self.command(f'DROP TABLE IF EXISTS {table}')
    except DbtDatabaseError:
        logger.info('Unexpected server exception dropping table')
```

The `try` wraps the whole `for` loop — if the first DROP raises (we see this in CH Cloud; the root cause of the DROP failure itself is unclear, possibly transient connection state after an EXCHANGE rejection), the second DROP never runs. Both probe tables leak.

PR #543 made the probe applicable to `Shared` engines (good), but didn't change the cleanup-loop shape or cache the probe result.

## Repro

Any dbt project against a CH Cloud workspace where `_check_atomic_exchange` raises during cleanup. We consistently observe `__dbt_exchange_test_*` tables persisting in `system.tables` after `dbt run` completes; the count grows by ~2 per affected connection per invocation.

## Suggested fix

Three independent improvements — any one helps; ideally all three:

1. **Per-iteration cleanup**: move the `try/except DbtDatabaseError` inside the `for` loop so a single DROP failure doesn't abort the rest.
2. **Cache the probe result on the adapter instance**: `_check_atomic_exchange` is deterministic per session, so re-running it many times per `dbt run` is wasted work.
3. **Optionally short-circuit on CH Cloud**: `SELECT value FROM system.settings WHERE name = 'cloud_mode'` returns `'1'` on Cloud; on Shared databases the EXCHANGE is known-supported, so the probe can be skipped entirely. (Less safe if CH Cloud ever changes that guarantee; (1)+(2) alone are sufficient to stop the leak.)

I lean toward (1) + (2). Happy to defer to your preference.

## How we'd like to help

If a fix is already in flight, happy to test against our Cloud workspace. Otherwise, glad to send a PR with whichever combination of (1)/(2)/(3) you prefer — please let us know the project's preferred direction.

Related symptom from a different code path: leaks of `<model>__dbt_new_data_<invocation_id>` tables from the `delete+insert` incremental strategy, filed separately.

Thanks for the great adapter.

29antonioac:
## Summary

Reduce connection startup overhead from the `EXCHANGE TABLES` capability check (`check_exchange`). Closes #641. Partially addresses #615.

**Context:** On every connection open, `_check_atomic_exchange` verifies that the server supports `EXCHANGE TABLES` by running up to 6 round-trips: `SELECT engine`, 2× `CREATE TABLE`, `EXCHANGE TABLES`, 2× `DROP TABLE`. On ClickHouse Cloud (`Shared` engine), these are not needed since `EXCHANGE` works. With multiple threads, they run in parallel for each connection — multiplying the load on the server for no benefit, since the result is identical for every connection within the same dbt run.

**Changes:**

1. **Shared engine short-circuit** (`_run_exchange_test`): `Shared` is a ClickHouse Cloud-only engine that guarantees `EXCHANGE TABLES` support. The function now returns `True` immediately after the engine `SELECT`, saving 5 DDL round-trips per connection.

2. **Process-level result caching** (`_check_atomic_exchange`): The test result is cached in a module-level variable behind a `threading.Lock`. Within a single dbt invocation, all connections share the cached result — the DDL test runs at most once regardless of thread count.

**This also fully resolves the orphan table leak described in #641.** On ClickHouse Cloud (Shared engine), `_check_atomic_exchange` previously created two `__dbt_exchange_test_*` MergeTree tables and left them behind if the cleanup `DROP` loop raised mid-iteration. With the Shared short-circuit, no test tables are ever created.

**Measured impact** on a ClickHouse Cloud (Shared) instance, running 2 view models against 9 dev schemas:

| | Wall time |
|---|---|
| Before | 4.28s |
| After (this PR) | 2.40s |

## Checklist

---

> [!NOTE]
> **Low Risk**
> Behavioral optimization only: same boolean outcome for atomic exchange, with less server load and no change to materialization SQL paths.
> 
> **Overview**
> Connection startup no longer repeats the full **`EXCHANGE TABLES`** probe on every open connection. The DDL path is moved into **`_run_exchange_test`**, while **`_check_atomic_exchange`** stores the outcome in a process-wide cache guarded by a lock so the test runs at most once per dbt process (including under multiple threads).
> 
> On ClickHouse Cloud (**`Shared`** database engine), the check returns **`True`** immediately after the engine **`SELECT`**, avoiding five extra DDL round-trips and the temporary **`__dbt_exchange_test_*`** tables. **CHANGELOG** documents the improvement; new unit tests cover Shared short-circuit, non-atomic engines, caching, and thread safety.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 965b3791531716671e9108c7a89ca520f5fe4a60. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Improvements
 * Starting with this release the `dbt-clickhouse` packages will be published to PyPI using Github Actions as a [Trusted Publisher](https://docs.pypi.org/trusted-publishers/). This will improve both the usability and the security of the release process ([#614](https://github.com/ClickHouse/dbt-clickhouse/pull/614)).
 * Populate `query_id` in `AdapterResponse` for every executed query. The query ID is generated as a UUID4 and forwarded to ClickHouse, making it available via `adapter_response` in dbt artifacts and enabling tools like Elementary to correlate dbt model runs with entries in `system.query_log`.
+* Reduce connection startup overhead from the `EXCHANGE TABLES` capability check. On ClickHouse Cloud (Shared engine), the check now short-circuits immediately after detecting the engine — skipping 5 DDL round-trips (2× `CREATE TABLE`, `EXCHANGE TABLES`, 2× `DROP TABLE`) that were previously run on every connection open. For all other engine types, the result is cached behind a process-level lock so the DDL test runs at most once per dbt invocation regardless of thread count. ([#653](https://github.com/ClickHouse/dbt-clickhouse/pull/653)).
 
 
 ### Release [1.10.0], 2026-02-16
--- a/dbt/adapters/clickhouse/dbclient.py
+++ b/dbt/adapters/clickhouse/dbclient.py
@@ -1,7 +1,8 @@
 import copy
+import threading
 import uuid
 from abc import ABC, abstractmethod
-from typing import Dict
+from typing import Dict, Optional
 
 from dbt.adapters.clickhouse.credentials import ClickHouseCredentials
 from dbt.adapters.clickhouse.errors import (
@@ -15,6 +16,9 @@ from dbt.adapters.clickhouse.query import quote_identifier
 from dbt.adapters.clickhouse.util import compare_versions, engine_can_atomic_exchange
 from dbt.adapters.exceptions import FailedToConnectError
 from dbt_common.exceptions import DbtConfigError, DbtDatabaseError
+
+_exchange_lock = threading.Lock()
+_exchange_result: Optional[bool] = None
 
 LW_DELETE_SETTING = 'allow_experimental_lightweight_delete'
 ND_MUTATION_SETTING = 'allow_nondeterministic_mutations'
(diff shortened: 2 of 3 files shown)


koletzilla (approved):
LGTM, thanks for the contribution!

PR #657, opened 2026-05-27
Add ClickHouse catalog relations support

dataders:
## Summary

- add `clickhouse__get_catalog_relations` for relation-scoped catalog metadata fetches
- mark `SchemaMetadataByRelations` as fully supported and route small filtered catalog requests through the relation-scoped path
- add unit coverage for the capability and dispatch choice

## Rationale

This follows the catalog fetch performance work introduced in the dbt Core adapter maintainer discussion: https://github.com/dbt-labs/dbt-core/discussions/8307#:~:text=Catalog%20fetch%20performance%20improvements

`get_catalog_relations()` lets an adapter fetch metadata for the exact relations dbt needs instead of scanning every object in every requested schema. This was useful but not mission critical for dbt Core adapters; it is more important in Fusion because catalog metadata lookup sits on a hotter interactive path where broad warehouse metadata scans are much harder for users to reason about.

Before this PR, dbt-clickhouse declared `SchemaMetadataByRelations` unsupported and used a schema-scoped catalog fetch plus Python post-filtering fallback. This switches to the simpler relation-scoped SQL macro, aligning dbt-clickhouse with the Fusion ClickHouse implementation in dbt-labs/fs#10553 and dbt-labs/dbt-fusion#2224.

## Validation

- `uv run --with-editable . --with-requirements dev_requirements.txt make lint`
- `uv run --no-project --python /Users/dataders/Developer/dbt-clickhouse/dev-env/.venv/bin/python python -m pytest tests/unit/test_catalog.py`
- `uv run --no-project --python /Users/dataders/Developer/dataface/.venv/bin/python python -m black --check dbt/adapters/clickhouse/impl.py tests/unit/test_catalog.py`
- `uv run --no-project python -m py_compile tests/unit/test_catalog.py dbt/adapters/clickhouse/impl.py`
- `git diff --check`

---

> [!NOTE]
> **Low Risk**
> Metadata-query routing and macro refactor only; large relation sets still fall back to schema-scoped catalog with existing Python filtering.
> 
> **Overview**
> Adds **relation-scoped catalog metadata** so dbt can query `system.columns` / `system.tables` for specific relations instead of whole schemas.
> 
> The adapter now advertises **`SchemaMetadataByRelations` as fully supported** and **`get_filtered_catalog`** uses **`get_catalog_by_relations`** when the relation set is present, within **`MAX_SCHEMA_METADATA_RELATIONS`**, and the capability is enabled; otherwise it keeps the schema-wide path. **`catalog.sql`** is refactored so **`clickhouse__get_catalog`** and the new **`clickhouse__get_catalog_relations`** share **`get_catalog_results_sql`**, with separate schema- vs relation-based `WHERE` builders (relations without a schema raise a compiler error).
> 
> Unit tests cover the capability flag and the relation-scoped dispatch; the changelog notes the feature.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3bd186092c080c2c536b3e7ad5888242a0d349c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/dbt/adapters/clickhouse/impl.py
+++ b/dbt/adapters/clickhouse/impl.py
@@ -76,7 +76,7 @@ class ClickHouseAdapter(SQLAdapter):
 
     _capabilities: CapabilityDict = CapabilityDict(
         {
-            Capability.SchemaMetadataByRelations: CapabilitySupport(support=Support.Unsupported),
+            Capability.SchemaMetadataByRelations: CapabilitySupport(support=Support.Full),
             Capability.TableLastModifiedMetadata: CapabilitySupport(support=Support.Unsupported),
         }
     )
@@ -419,7 +419,15 @@ class ClickHouseAdapter(SQLAdapter):
         used_schemas: FrozenSet[Tuple[str, str]],
         relations: Optional[Set[BaseRelation]] = None,
     ):
-        catalog, exceptions = self.get_catalog(relation_configs, used_schemas)
+        if (
+            relations is None
+            or len(relations) > self.MAX_SCHEMA_METADATA_RELATIONS
+            or not self.supports(Capability.SchemaMetadataByRelations)
+        ):
+            catalog, exceptions = self.get_catalog(relation_configs, used_schemas)
+        else:
+            catalog, exceptions = self.get_catalog_by_relations(used_schemas, relations)
+
         if relations and catalog:
             relation_map = {(r.schema, r.identifier) for r in relations}
 
--- a/dbt/include/clickhouse/macros/catalog.sql
+++ b/dbt/include/clickhouse/macros/catalog.sql
@@ -1,26 +1,18 @@
 {% macro clickhouse__get_catalog(information_schema, schemas) -%}
   {%- call statement('catalog', fetch_result=True) -%}
-    {{ clickhouse__get_catalog_schemas_sql(information_schema, schemas) }}
+    {{ get_catalog_results_sql(get_catalog_schemas_where_clause_sql(schemas)) }}
   {%- endcall -%}
   {{ return(load_result('catalog').table) }}
 {%- endmacro %}
 
 {% macro clickhouse__get_catalog_relations(information_schema, relations) -%}
   {%- call statement('catalog', fetch_result=True) -%}
-    {{ clickhouse__get_catalog_relations_sql(information_schema, relations) }}
+    {{ get_catalog_results_sql(get_catalog_relations_where_clause_sql(relations)) }}
   {%- endcall -%}
   {{ return(load_result('catalog').table) }}
 {%- endmacro %}
 
-{% macro clickhouse__get_catalog_schemas_sql(information_schema, schemas) -%}
-  {{ clickhouse__get_catalog_results_sql(clickhouse__get_catalog_schemas_where_clause_sql(schemas)) }}
-{%- endmacro %}
-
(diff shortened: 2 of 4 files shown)


dbt/include/clickhouse/macros/catalog.sql
```diff
@@ -1,5 +1,26 @@
...
+  {{ return(load_result('catalog').table) }}
+{%- endmacro %}
+
+{% macro clickhouse__get_catalog_schemas_sql(information_schema, schemas) -%}
+  {{ clickhouse__get_catalog_results_sql(clickhouse__get_catalog_schemas_where_clause_sql(schemas)) }}
+{%- endmacro %}
+
+{% macro clickhouse__get_catalog_relations_sql(information_schema, relations) -%}
+  {{ clickhouse__get_catalog_results_sql(clickhouse__get_catalog_relations_where_clause_sql(relations)) }}
+{%- endmacro %}
```
koletzilla: Looks like we can get rid of these two macros.

for example `clickhouse__get_catalog` calls `clickhouse__get_catalog_schemas_sql` that then calls `clickhouse__get_catalog_results_sql`. we can just remove this level of indirection and just make `clickhouse__get_catalog` call `clickhouse__get_catalog_results_sql` with the appropriate arg.

same thing for `clickhouse__get_catalog_relations_sql`

dbt/include/clickhouse/macros/catalog.sql
```diff
@@ -1,5 +1,26 @@
...
+
+{% macro clickhouse__get_catalog_results_sql(where_clause) -%}
```
koletzilla: Can we leave the rest of these macros, that are only called from inside this macro, without the `clickhouse__` prefix in the name?

I'm not 100% if this is correct, but we are keeping the `clickhouse__` prefix for those macros that are used by the dbt macro dispatcher and just remove the prefix for those macros that are expected to be "private"

dbt/adapters/clickhouse/impl.py
```diff
@@ -419,7 +419,15 @@ def get_filtered_catalog(
         used_schemas: FrozenSet[Tuple[str, str]],
         relations: Optional[Set[BaseRelation]] = None,
     ):
-        catalog, exceptions = self.get_catalog(relation_configs, used_schemas)
+        if (
+            relations is None
+            or len(relations) > self.MAX_SCHEMA_METADATA_RELATIONS
+            or not self.supports(Capability.SchemaMetadataByRelations)
```
koletzilla, line 425: This the check for `Capability.SchemaMetadataByRelations: CapabilitySupport(support=Support.Full),` seems not needed now as we are declaring it as supported.

Is it because it's now more close to the original?

TBH I'd love to remove this and keep the original, but I see that the original do a casefold on the names of the tables and these names are case sensitive in ClickHouse.

Thinking out loud: If we could add a "case sensitivity" setting in dbt-core we may be able to just use the original function 🤔

dbt/include/clickhouse/macros/catalog.sql
koletzilla: applied

dbt/include/clickhouse/macros/catalog.sql
koletzilla: applied

koletzilla (approved):

Issue #372, opened 2024-10-28
Unnecessary Setting added


Shlomixg:
### Describe the bug
An unnecessary setting `replicated_deduplication_window=0` is added to all tables created using DBT:

```sql
{{
    config(
        materialized = 'table',
        schema = 'default_schema'
    )
}}

select 1
```

and result is:
```sql
-- default_schema.test_table definition

CREATE TABLE default_schema.test_table
(
    `1` UInt8
)
ENGINE = MergeTree
ORDER BY tuple()
SETTINGS replicated_deduplication_window = 0,
 index_granularity = 8192;
```

How can I disable it?

### Expected behaviour
Create the tables without the setting

### dbt and/or ClickHouse server logs
```sql
/* {"app": "dbt", "dbt_version": "1.8.8", "profile_name": "default_schema", "target_name": "default_schema", "node_id": "model.ra_dbt.test_table"} */
create table `datricks`.`test_table`
engine = MergeTree()
 order by (tuple())
-- end_of_sql
SETTINGS  replicated_deduplication_window=0
empty
as (
select 1
)
```

#### Environment
* dbt version: 1.8.8
* dbt-clickhouse version: 1.8.4
* clickhouse-driver version: 0.2.9
* Python version: 3.11.9
* Operating system: Debian 12 (on docker)
* We are not using Clickhouse Keeper\ZooKeeper or any distributed DDL

genzgd:
It is necessary in some cases to prevent duplicates.  Why do you need to disable it?

Shlomixg:
According to the [docs ](https://clickhouse.com/docs/en/operations/settings/merge-tree-settings#replicated-deduplication-window) this setting is relevant only to replicated tables on ClickHouse Keeper, which we are not using.
I don't want to disable it, just think it's not necessary to add it to each table

vik1109:
replicated_deduplication_window do not work with set() engine. How can i disable this setting?

babaMar:
> It is necessary in some cases to prevent duplicates. Why do you need to disable it?

Shouldn't then the user set it if they see duplicates? It looks like causing records to not being inserted in some cases.

koletzilla:
> replicated_deduplication_window do not work with set() engine. How can i disable this setting?
@vik1109 Thanks for commenting it, we fixed it in the release https://github.com/ClickHouse/dbt-clickhouse/releases/tag/v1.9.3

babaMar:
> Some related code:
> 
[...]

PR #677, opened 2026-06-30
fix: Avoid injected settings to reach manifest and overselect on `state:modified`

### LINKED ISSUE #676: `replicated_deduplication_window=0` is injected into manifest, breaking deferred runs
### Describe the bug

`get_model_settings` passes the `settings` dict from `model['config']` directly into `update_model_settings`, which mutates it in-place by injecting `replicated_deduplication_window='0'`. Because dbt serialises `manifest.json` after `dbt run`, this injected key gets persisted into the manifest.

The consequence is that commands like `dbt parse` or `dbt ls` — which never call `get_model_settings` — produce a clean manifest without the injected key. When using `--select state:modified+ --defer --state <prev>`, dbt compares the two manifests and sees `config_changed` on every affected model, even when nothing actually changed.

Related to #372 — that issue covers the unwanted DDL injection; this reports the deeper consequence of the in-place mutation leaking into the manifest.

### Steps to reproduce

```python
{{ config(
    materialized='table',
    engine='MergeTree()',
    order_by='id',
    settings={'allow_nullable_key': 1}
) }}
select 1 as id
```

1. `dbt run --select my_model` — produces `target/manifest.json` with `replicated_deduplication_window='0'` inside `config.settings`
2. `mkdir state && cp target/manifest.json state/`
3. `dbt run --select state:modified+ --defer --state state/` with no code changes

### Expected behaviour

Step 3 selects zero models — nothing changed. Actual: `my_model` is selected and re-run because dbt sees `config_changed` (the injected key is present in the state manifest but absent from the freshly parsed one).

The diff between the manifest produced by `dbt run` vs `dbt parse` for the same model:

```json
// state/manifest.json  (produced by dbt run)
"config": {
  "settings": {
    "allow_nullable_key": 1,
    "replicated_deduplication_window": "0"   ← injected at runtime, leaked into manifest
  }
}

// target/manifest.json  (produced by dbt parse / dbt ls)
"config": {
  "settings": {
    "allow_nullable_key": 1
                                              ← key absent
  }
}
```

In a project with 12 affected models this caused their entire downstream graph to be re-run on every CI pipeline, triggered by an unrelated change (e.g. editing a documentation `.yml` file).

### Fix

In `dbt/adapters/clickhouse/impl.py`, deep-copy the settings dict before passing it to `update_model_settings` so the in-memory model config is never mutated:

```python
# Before
settings = model['config'].get('settings', {})

# After
settings = copy.deepcopy(model['config'].get('settings', {}))
```

### dbt and/or ClickHouse server logs

Not applicable — the bug is silent. No warnings or errors are emitted; models are simply re-run unnecessarily on every deferred pipeline run.

#### Environment

* dbt version: 1.10.15
* dbt-clickhouse version: 1.10.1
* clickhouse-connect version (if using http): 0.15.1
* Python version: 3.13.13
* Operating system: macOS

#### ClickHouse server

* ClickHouse Server version: 26.2.1.455
* ClickHouse Server non-default settings: none relevant
* Engine of affected tables: `SharedMergeTree` (ClickHouse Cloud)

29antonioac:
## Summary

Fixes a bug where `get_model_settings` mutated `model['config']['settings']` in-place, causing adapter-injected settings (e.g. `replicated_deduplication_window='0'`) to be persisted into `manifest.json` after `dbt run`. This produced a permanent asymmetry with manifests generated by `dbt parse` or `dbt ls`, which never call `get_model_settings` and therefore produced a clean manifest. The mismatch caused `--select state:modified+ --defer` to report `config_changed` on every affected model on every subsequent run, even when nothing had changed.

Fix: deep-copy the settings dict in `get_model_settings` before passing it to `update_model_settings`, so the original `model['config']` is never touched.

Related to #372.
Closes #676

## Checklist


koletzilla:
Ey @29antonioac , thanks as always for reporting it and for sharing a PR 🙇 

I was wondering that the solution feels like should go in the other direction: Instead of skipping it, any setting the adapter add should be recorded in the manifest in both situations.

Out of curiosity, have you checked it we can do it otherwise? Maybe we need to change when this setting is added to make sure it's always included

29antonioac:
Hey mate, no problem! Happy to support OSS 😄 

Please take my opinion with a pinch of salt: I'm not a dbt expert or power user, just a guy trying to help his colleagues, I feel more comfortable with ClickHouse than dbt in general, so I can be missing details 😄 . 

I was wondering the same, maybe the manifest should get all settings of the model, including the injected ones. However, when finding the issue because of `replicated_deduplication_window=0`, my first thought was adapter-specific settings should not belong to the manifest: if tomorrow this setting disappears from the adapter (ClickHouse default, table engine refined behaviour or any other reason), all users will find their models flagged as modified when they didn't do any explicit change in the model. 

The setting only applies to MergeTree tables, not sure if `dbt parse` could retrieve this without a connection (I assume it can get the engine from the model). And given the code uses `conn` I guess the injection could depend on the server/instance?

I'm happy to look for a way to get the manifest with whatever setting the adapter injects at `dbt parse` step, as you are the main maintainers I trust you more than myself 😂  .

koletzilla:
> if tomorrow this setting disappears from the adapter (ClickHouse default, table engine refined behaviour or any other reason), all users will find their models flagged as modified when they didn't do any explicit change in the model.

Agree 100% that it wouldn't be nice at all, but after giving a second thought, maybe it's worth it to flag them 🤔. If some config change, like `replicated_deduplication_window`, you'd probably need to make sure you're not depending on that setting at all or your data will be impacted in the next run. I guess most of the settings would be probably false positives as you mentioned: things that are now default, settings that are no longer needed... but worth checking in case you're still running on an old system and maybe these settings still apply to you 🤔 

Would you check if there's indeed a way to do that? Btw, I thought version checking may get in the middle of it, so assume in your checks that the version checking is removed. I already removed that code here https://github.com/ClickHouse/dbt-clickhouse/pull/678/changes

29antonioac:
@koletzilla It's not the nicest solution, but with Claude and inspecting dbt core it seems there is a point where the manifest is resolved at parse time and we can inject settings there instead: `set_macro_resolver`. The solution looks clean to me, injecting this in just one place and being able to have the same one after parsing and after running.

It feels a bit hacky though. Also the solution in the new test to read explicitly the manifest without `get_manifest` makes me feel a bit uncomfortable 😂 . Not sure if we are smashing parts of the dbt core flow.

I removed the version comparison too as you mentioned 😄

PR #678, opened 2026-07-03
Simplyfing dbt-clickhouse codebase to make it easier to develop in dbt Core 2.0

koletzilla:
# Summary

Compatibility fixes for dbt Core v2 so macros in both places keep synced:

- **`?` in regexes replaced with `\x3F`** (`adapters.sql`, `materialized_view.sql`): v2's driver treats a literal `?` in query text as a bind parameter, breaking the regexes that extract MV target tables. `\x3F` is the ClickHouse string-literal escape for the same character, so behavior is identical. Can be reverted once [adbc_clickhouse#53](https://github.com/ClickHouse/adbc_clickhouse/issues/53) is fixed.
- **`modules.re.findall(..., flags=...)` → positional arg** (`materialized_view.sql`): v2's `re` module doesn't support the `flags=` keyword.
- **`primary_key` config now accepts lists** (`table.sql`): v2's typed configs listify scalar values, so the validator now allows `list`, lists are joined into a parenthesized expression (bare `PRIMARY KEY a, b` is a ClickHouse syntax error), and empty values (`''`/`[]`) omit the clause instead of emitting invalid DDL. Covered by a new integration test (`TestTableMaterializationPrimaryKey`) exercising string, list, and empty-list configs.
- **Catalog macros renamed with `clickhouse__` prefix** (`catalog.sql`): the unprefixed helper names collided with dbt-core's own macros under v2's dispatch. Also `null as table_owner` → `cast(null as Nullable(String))` for stricter typing.
- **`toString(access_type)` in the show-grants query** (`apply_grants.sql`): v2 needs the enum returned as a string.

## Dead code removal

- **Dropped pre-22.7 ClickHouse support paths**: removed the server-version check around the deduplication-window setting in `dbclient.py` (and its now-unused `compare_versions` import), and the pre-21.9 heredoc version guard in `persist_docs.sql`. These versions are long EOL.

## Cosmetic

- Trailing newlines added / blank lines removed in several macro files (`snapshot.sql`, `timestamps.sql`, `is_incremental.sql`, etc.) so the files match v2's copies byte-for-byte.


dbt/adapters/clickhouse/impl.py
```diff
@@ -150,14 +150,6 @@ def clickhouse_db_engine_clause(self):
             return f'ENGINE {conn.credentials.database_engine}'
         return ''
 
-    @available
-    def is_before_version(self, version: str) -> bool:
```
BentsiLeviav, line 153: Isn't that a useful functionality to have in a project? 
I understand that the changes in this PR make this function/logic redundant, but generally, it sounds plausible to have version-based checks.

koletzilla:
PR moved back to draft as it's WIP

dbt/adapters/clickhouse/impl.py
koletzilla: 100% agree on this functionality may be useful in the future, but keeping these now-unused functions is adding a bit of noise so I'd like to get rid of it.

I'm encountering many situations were I can't just "copy-paste" logic from dbt-clickhouse to dbt-fusion and that logic is making all things harder to move as I need to figure out how these things should be done in Fusion now.

I prefer to go more lean now: Remove what is unused so I'm faster moving real features to dbt Core. If in the future we need them again, we can just go back in git and copy the code.

I have changed this PR back to a draft as I'm encountering other things that can be simplified. I'll add them here and put it back to a regular PR once I have a nice set of features to review.

koletzilla: I'm finally not going to remove the base methods. We are going to need them for https://github.com/ClickHouse/dbt-clickhouse/pull/680

koletzilla:
bugbot run

koletzilla:
Pr ready to review! I have added to the description all the changes I have done, but basically all are related to changes done to make the code compatible with V2

mzitnik (approved):
LGTM

PR #680, opened 2026-07-07
Fix replace partition error from version >26.6

koletzilla:
## Summary
Fix the `insert_overwrite` incremental strategy failing on ClickHouse 26.6+ with `Code: 36 ... refusing REPLACE PARTITION because it would silently drop the destination partition's data`. The strategy intentionally replaces partitions from a source table that may have no parts for a partition (to clear shards that received no new data on distributed tables), which newer servers reject by default. The `REPLACE PARTITION` statement now includes `allow_replace_partition_from_empty_source=1` on servers 26.6 and newer; older servers keep their existing behavior.

I went with the simplest solution: Just add the setting for this specific use case as we explicitly want this one to behave this way. Not touching other queries in case we want to have a different behaviour for them.

Avoided other possible implementations with the focus to keep it simple:
- Checking if the setting is available by checking existing settings: Just checking by version is enough.
- Adding it to the list of current settings applied: Many required changes to do that for just this one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### Unreleased
+
+#### Bug Fixes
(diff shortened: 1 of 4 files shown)


koletzilla:
test_cloud error is another flaky test. Solving it here: https://github.com/ClickHouse/dbt-clickhouse/pull/681

BentsiLeviav (approved):
LGTM

PR #683, opened 2026-07-14
Fix multi-mv marker extraction bug

### LINKED ISSUE #682: Multi-MV marker extraction bug: leading `--` comment silently blanks the view query because `[^:]+` regex matches newlines
### Describe the bug

The `materialized_view` materialization extracts each MV's SELECT from the compiled SQL via `-- <name>:begin` … `-- <name>:end` markers. The view-name capture on line 182 is `[^:]+`, which matches newlines:

See https://github.com/ClickHouse/dbt-clickhouse/blob/main/dbt/include/clickhouse/macros/materializations/materialized_view.sql#L182

```jinja
{% set view_names = modules.re.findall('--(?:\s)?([^:]+):begin', sql) %}
```

If any `--` comment precedes the first `:begin` with no `:` between them, the regex anchors on that earlier `--` and `([^:]+)` greedily consumes across newlines up to the real marker's colon. Example:

```sql
-- added to fix bug #1267
-- mv:begin
select ....
-- mv:end
```

Then line 188 plugs that captured name back in to pair begin<->end:

```jinja
{% set view_sql = modules.re.findall('--(?:\s)?' + view_name + ':begin(.*)--(?:\s)?' + view_name + ':end', sql, flags=modules.re.DOTALL)[0] %}
```

The garbage name matches on the `:begin` side (bc that's where it was carved from) but not on `:end` (the real text there is just `-- mv:end`). So the pair-up fails, `findall(...)` returns [], and [][0] renders empty. So dbt emits `create materialized view … as / alter table … modify query` with an empty query body.

The bug is masked whenever the leading comment contains a `:` (the `:` stops `[^:]+`), which is why canonical usage works. With multiple MVs, the first marker is swallowed and dropped while later ones extract correctly, so an MV can go silently missing.

Fix: constrain the name to a bare token: `([^:\s]+)` on line 182. This still matches every valid `-- name:begin` and a stray `--` can no longer span into the marker.

### Steps to reproduce

1. Create a materialized_view model using the multi-MV markers, with a leading `--` comment that has no colon before the first `:begin` (see example above).
2. dbt run.
3. ClickHouse rejects the empty query.

### Expected behaviour

The view name is a bare token; an unrelated leading -- comment must not leak into the marker. The select between `-- mv:begin / -- mv:end` is extracted as the MV query.

Example model:
```sql
{{ config(materialized='materialized_view', engine='Null()') }}
-- publishes rows to the target table
-- mv:begin
select 1 as x
-- mv:end
```

Regex in isolation:

```python
import re
PAT = r'--(?:\s)?([^:]+):begin'
re.findall(PAT, "-- mv:begin\nx\n-- mv:end")                    # ['mv']                    OK
re.findall(PAT, "-- publishes rows\n-- mv:begin\nx\n-- mv:end") # ['publishes rows\n-- mv'] BROKEN
```

dbt and/or ClickHouse server logs

```
Code: 62. DB::Exception: Syntax error: failed at position N (end of query): Expected one of: SELECT query, WITH, FROM, SELECT. (SYNTAX_ERROR)
```

#### Environment
- dbt version: 1.10.20
- dbt-clickhouse version: 1.10.0 (also reproduces on main, fa83906)
- clickhouse-connect version (if using http): 0.15.0
- Python version: 3.12

jivesh-allium:
## Summary

Fixes #682.

Writing multiple MVs in a single file breaks when a `--` comment appears before the first `--name:begin` marker. The view-name capture on line 182 of `materialized_view.sql` is `[^:]+`, which matches newlines, so a leading comment (with no colon before the marker) is greedily pulled across newlines into the marker. The captured name is corrupted, the begin/end pair-up fails, and dbt emits an empty view query.  This results in either a ClickHouse `SYNTAX_ERROR` or a silently-dropped first MV when several are defined.

This PR constrains the name capture to a bare token (`[^:\s]+`), so a stray `--` can no longer span into the marker. Adds a regression test covering a model with a leading no-colon comment before the markers.

## Checklist


koletzilla:
Thanks for the PR @jivesh-allium .

One thing: Can you please change the test so instead of a full integration one? I feel we'd like at some moment to test more variations instead of just the happy path + the error you are fixing.

We should isolate the part of the macro that searchs for the `view_names` and generates the list of `views` to its own macro. Then test it with at least two variations: the happy path + the path you fixed.

I think we don't have right now a way to unit-test macros. We should create some generic utility in our unit-tests so we can easily test them. That would be part of the task I'm asking.

That way we can have good testing for that + make easier to test more possible variations in the future.

Would you do that? Thanks!

jivesh-allium:
Thanks for the review @koletzilla, your suggestion makes sense!

I've pulled the marker parsing into its own `clickhouse__extract_mv_views` macro, and added a `MacroHarness` in `tests/unit/macro_harness.py` that calls macros by name through dbt's Jinja environment. `test_extract_mv_views.py` uses it and replaces the integration test I'd added before.

caveat on "generic": the harness currently provides `return()`, `modules`, `log()` and `exceptions`, anything else gets injected per call. Importantly a macro using `{% call statement(...) %}` needs a fake `statement` supplied, which we can add as a test helper in a follow-up.

Let me know if the utility looks useful, and whether there's a test organization direction you'd prefer, for example a dedicated directory for macro tests as more get added.

koletzilla:
Thanks for the change! Will merge it after tests pass

Issue #221, opened 2023-12-12
Atomic recreation of views


krivars:
### Describe the bug
When views are recreated, it is not atomic. The run might fail after the existing view has been renamed to `backup`, but before the new (intermediate) has been renamed to the target.
https://github.com/ClickHouse/dbt-clickhouse/blob/main/dbt/include/clickhouse/macros/materializations/view.sql#L42

### Steps to reproduce
1. `dbt run -m some_view_that_already_exists`
2. If it fails between line 47 and 48 in the code linked above, your don't have your view anymore.

### Expected behaviour
I though it would be possible with atomic view re-creation. It is possible to run `CREATE OR REPLACE VIEW`, or is that not an atomic operation?

genzgd:
Actually I'm wrong, there is `REPLACE VIEW` (I'm not sure it's atomic), I was thinking of tables and cluster operations.  We can look using that instead.

genzgd:
It turns out the REPLACE VIEW is not, in fact, atomic, so changing the DDL won't guarantee a fix for your particular issue. 
Regardless, it's definitely cleaner and worth doing as time permits.  This could be fixed eventually in ClickHouse server.

In general however, I would not want to construct workflows that depend on atomic operations in ClickHouse via dbt.  True atomic operations especially across a ClickHouse cluster are extremely difficult to guarantee, and dbt itself isn't going to be aware of various temporary tables and other mechanisms across a cluster either.

Issue #340, opened 2024-08-16
Running DBT on MacOS with dbt-clickhouse does not support renameat2()


LPauzies:
### Describe the bug
When running : `dbt run --target=clickhouse-dev --select=dbt_mart` on MacOS Sonoma 14.6.1

```
08:27:45  Running with dbt=1.7.17
08:27:46  Registered adapter: clickhouse=1.7.7
08:27:46  Found 65 models, 3 seeds, 280 tests, 85 sources, 0 exposures, 0 metrics, 942 macros, 0 groups, 0 semantic models
08:27:46  
08:27:48  dbt_clickhouse adapter: ClickHouse server does not support the EXCHANGE TABLES command
08:27:48  dbt_clickhouse adapter: This can be caused by an obsolete ClickHouse version or by running ClickHouse on
08:27:48  dbt_clickhouse adapter: an operating system that does not support the low level renameat2() system call.
08:27:48  dbt_clickhouse adapter: Some DBT materializations will be slower and not atomic as a result.
08:27:49  dbt_clickhouse adapter: Unexpected server exception dropping table
08:27:49  
08:27:49  Finished running  in 0 hours 0 minutes and 2.68 seconds (2.68s).
08:27:49  Encountered an error:
Database Error
  'NoneType' object has no attribute 'version_major'
make: *** [dbt-run] Error 2
```

With : 
- clickhouse : 23.8.4.69
- dbt-clickhouse : 1.7.7
- dbt-core : 1.7.17

While the same code is running perfectly in Linux distributions.

### Steps to reproduce
1. Get a MacOS Sonoma 14.6.1
2. Create a DBT project
3. Create a clickhouse-server instance with version 23.8.4.69
4. Create a connection profile for this instance called "my_profile"
5. Create a dummy model called "my_mart" in the DBT project
6. Execute `dbt run --target=my_profile --select=my_mart`

### Expected behaviour
A correct DBT materialization

### Code examples, such as models or profile settings
profiles.yml
```
clickhouse-dev:
      type: clickhouse
      driver: native
      secure: true
      verify: false
      schema: dbt
      retries: 3
      connect_timeout: 100
      send_receive_timeout: 3000
      sync_request_timeout: 15
      threads: 4
      host: localhost # SSH tunneling
      port: 9430  # ssh port mapping
      user: user
      password: password
      custom_settings: {max_bytes_before_external_group_by: '68719480000', max_bytes_before_external_sort: '68719480000'}
```

[...]

PR #684, opened 2026-07-16
fix: fall back to drop + create when replacing views without renameat2 support

nimast:
## Summary

On filesystems without renameat2(RENAME_EXCHANGE) support (NFS, EFS, CephFS, older ZFS, some Docker bind mounts), every view model fails on the second dbt run with:

Code: 1. DB::Exception: System call renameat2() is not supported. (UNSUPPORTED_METHOD)

The first run succeeds because CREATE OR REPLACE VIEW on a nonexistent target takes a plain-rename path; only replace-over-existing requires the exchange syscall. ClickHouse has no server-side fallback for this in any version, and a fix is not planned (see https://github.com/ClickHouse/ClickHouse/issues/96835) so the adapter has to handle it.

Tables and incremental models already fall back to non-atomic operations when the connection reports that atomic exchange is unsupported (#114); views never did — the view materialization emits create or replace view unconditionally (since #412). 

The fix: when replacing an existing view and adapter.supports_atomic_exchange() is false, drop the view first so the subsequent create or replace view takes the plain-rename path. When exchange is supported, behavior is unchanged — the atomic replace from #221 is preserved.

The test: a regression test that emulates the filesystem limitation with a seccomp profile making renameat2 return ENOSYS inside a dedicated ClickHouse container, then runs a view model twice against it. Verified it reproduces the exact UNSUPPORTED_METHOD failure without the fix and passes with it (also verified against a real environment where EXCHANGE TABLES genuinely fails). The test skips cleanly when Docker is unavailable.

Note for reviewers: check_exchange: false in a profile skips the probe and assumes exchange works, so it disables this fallback — users hitting this issue should keep the default check_exchange: true.

## Checklist


koletzilla:
Hi! Thanks for the PR.

I think this PR is not dealing with #340 but with a different issue. #340 is about the actual inner error (exception: Code: 497. DB::Exception: user: Not enough privileges.) not showing up. Your PR seems like an actual renameat2 problem. Can you confirm that?

koletzilla:
About the issues in test_local_matrix: the PR needs to be rebased on top of latests main as it contains the fix. Dont worry about test_cloud, it fails in every contributor PR

koletzilla:
bugbot run

nimast:
> Hi! Thanks for the PR.
> 
> 
> 
> I think this PR is not dealing with #340 but with a different issue. #340 is about the actual inner error (exception: Code: 497. DB::Exception: user: Not enough privileges.) not showing up. Your PR seems like an actual renameat2 problem. Can you confirm that?
> 
> 

hey sure thing, yes this is an actual renameat2 issue, sorry about that, not sure how the other one got mentioned. Were running the patched version without issues. It appears when Clickhouse is run in Kubernetes over OpenShift with NFS backing storage. The upstream issue in Clickhouse is a 'will not fix' as there appears go be no atomic solution for this filesystem, but the adapter change seems safe.

koletzilla (approved):

PR #688, opened 2026-07-21
 Update refreshable MV parameters on dbt run via MODIFY REFRESH

### LINKED ISSUE #611: Update refreshable MV's parameters on `dbt run`
**Is your feature request related to a problem? Please describe.**
Once a materialized view is created, each subsequent `dbt run` operation updates its SQL using the `ALTER TABLE MODIFY QUERY` operation, enabling changes without downtime. It would be beneficial to also provide a way to update the refresh parameters during `dbt run`.

**Describe the solution you'd like**
If it's a refreshable materialized view, we can also update its parameters without downtime using the `ALTER TABLE [db.]name MODIFY REFRESH EVERY|AFTER ... [RANDOMIZE FOR ...] [DEPENDS ON ...] [SETTINGS ...]` operation as described in https://clickhouse.com/docs/sql-reference/statements/create/view#changing-refresh-parameters

**Describe alternatives you've considered**
Currently, to update the refresh parameters, we need to execute `dbt run --full-refresh` on the materialized view. This process drops and recreates it, which causes several issues:
* A refresh may start immediately after the recreate, which may be unnecessary.
* Internally, the reference in `system.view_refreshes` may now point to a different view.

BentsiLeviav:
## Summary

Closes #611

Until now, changing the refresh parameters of a refreshable MV (things like interval, randomize, depends_on) required a `--full-refresh` run, which drops and recreates the view. This PR applies them in place on every dbt run using `ALTER TABLE ... MODIFY REFRESH`, the same way we already update the query via `MODIFY QUERY` - no downtime, no unwanted refresh on recreate, and the view keeps its identity in `system.view_refreshes`.

Some transitions can't be done in place (without querying CH for additional metadata), and previously they were silently ignored. They now fail with an error pointing the user to `--full-refresh`:

- turning an existing regular MV into a refreshable one (or the other way around)
- changing the append setting

To detect these cases without extra queries, the deployed state (`is_refreshable`, `refreshable_append`) is parsed from the DDL we already fetch in `list_relations_without_caching` and carried on the cached relation.

## Checklist

### DIFF
--- a/dbt/include/clickhouse/macros/adapters.sql
+++ b/dbt/include/clickhouse/macros/adapters.sql
@@ -51,6 +51,10 @@
         map('schema', mv_sources.mv_database, 'name', mv_sources.mv_name, 'sql', mv_sources.mv_sql),
         mv_sources.mv_name != ''
       ) as mvs_pointing_to_it,
+      -- The refresh clause of a refreshable MV sits between the view name and the TO clause,
+      -- e.g. CREATE MATERIALIZED VIEW db.mv REFRESH EVERY 2 MINUTE [APPEND] TO db.target ...
+      max(position(substring(t.create_table_query, 1, position(t.create_table_query, ' TO ')), ' REFRESH ')) > 0 as is_refreshable,
+      max(position(substring(t.create_table_query, 1, position(t.create_table_query, ' TO ')), ' APPEND')) > 0 as refreshable_append,
       {%- if adapter.get_clickhouse_cluster_name() -%}
         count(distinct _shard_num) > 1  as  is_on_cluster
         from clusterAllReplicas({{ adapter.get_clickhouse_cluster_name() }}, system.tables) as t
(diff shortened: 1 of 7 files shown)


dbt/include/clickhouse/macros/adapters.sql
```diff
@@ -51,6 +51,10 @@
         map('schema', mv_sources.mv_database, 'name', mv_sources.mv_name, 'sql', mv_sources.mv_sql),
         mv_sources.mv_name != ''
       ) as mvs_pointing_to_it,
+      -- The refresh clause of a refreshable MV sits between the view name and the TO clause,
+      -- e.g. CREATE MATERIALIZED VIEW db.mv REFRESH EVERY 2 MINUTE [APPEND] TO db.target ...
+      max(position(substring(t.create_table_query, 1, position(t.create_table_query, ' TO ')), ' REFRESH ')) > 0 as is_refreshable,
```
BentsiLeviav, line 56: The model config only knows the desired state. The two flags exist to know the deployed state without a per-MV query

koletzilla:
LGTM! Merging it

koletzilla (approved):

PR #689, opened 2026-07-22
fix: apply query_settings during contract schema introspection

### LINKED ISSUE #667: Contract not respect query_settings
# Bug: Contract validation fails when `join_use_nulls` is configured via `query_settings`

## Description

When `join_use_nulls=1` is configured through `query_settings` in a model config, dbt contract validation incorrectly fails for a nullable column.

The same model passes contract validation when `join_use_nulls` is specified directly in the query instead.

### `my_model.sql`

```sql
{{
    config(
        order_by='id',
        query_settings={
            'join_use_nulls': 1
        }
    )
}}

select
    id,
    last_value
from (
    select 1 as id
) t1
left join (
    select
        1 as id,
        max(value) as last_value
    from (
        select 1 as value
    )
    where value = 2
) t2
    on t1.id = t2.id
```

### `my_model.yml`

```yaml
models:
  - name: my_model
    config:
      contract:
        enforced: true
    columns:
      - name: id
        data_type: UInt8
      - name: last_value
        data_type: Nullable(UInt8)
```

## Expected behavior

Contract validation should succeed because `last_value` is returned as `Nullable(UInt8)` when `join_use_nulls=1` is enabled.

## Actual behavior

Contract validation fails due to a type mismatch.

## Additional information

If `join_use_nulls` is specified directly in the model query instead of through `query_settings`, contract validation succeeds.

## Environment

* dbt version: 1.10.19
* dbt-clickhouse version: 1.10.0
* ClickHouse version: 25.4

sankalpsthakur:
## Summary

Fixes #667

Contract validation (`clickhouse__get_assert_columns_equivalent`) and schema-change introspection (`check_incremental_schema_changes`) called `get_column_schema_from_query` without the model's `query_settings`. Settings such as `join_use_nulls=1` therefore did not apply when deriving column types, so a contract that correctly declares `Nullable(...)` for left-join results would fail validation even though the runtime query (which does apply `query_settings`) produces those types.

### Changes
- Pass `query_settings` into `get_column_schema_from_query` (client-level settings on the `SELECT * FROM (...) LIMIT 0` introspection query).
- Wire `query_settings` through contract assertion, incremental/MV schema-change checks, and distributed empty-table column inference.
- Forward kwargs (including `settings`) in the native client's `columns_in_query`.
- Unit tests for settings forwarding; integration regression matching the issue repro.

Independent of #690 (MV `catchup=false` projection/index DDL); both are dbt-clickhouse bugfixes from the same batch.

## Checklist


koletzilla:
bugbot run

koletzilla:
LGTM! Thanks for the PR :D

koletzilla (approved):

Issue #616, opened 2026-02-19
Raise an exception/warning when `repopulate_from_mvs_on_full_refresh` and `catchup` are set


BentsiLeviav:
### Description

Our documentation states that

> Materialized views may be configured with `catchup=False` or the target table may be configured with `repopulate_from_mvs_on_full_refresh=False`, so no backfill is executed when the materialized views are created or when the target table is recreated. This is the expected behavior, so if you want to reinsert the data using the materialized views SQL, make sure to set `catchup=True` in the materialized view (this is the default value) or `repopulate_from_mvs_on_full_refresh=True` in the target table. Make sure you are not activating both at the same time to avoid duplicates. Check the [configuration section](#explicit-target-configuration) for more details.

We might want to explore adding a check and raising an exception or a warning if these two settings are set. 

see https://github.com/ClickHouse/clickhouse-docs/pull/5534#discussion_r2816030958 for referance

PR #690, opened 2026-07-22
fix: create projections/indexes when catchup=false on MVs

### LINKED ISSUE #637: catchup=false silently drops projections and indexes config on first-create / --full-refresh
### Describe the bug

When a **materialized_view** model is configured with both `catchup=false` and `projections=[...]` (or indexes=[...]), the projections/indexes are silently not created on the target table during first materialization or --full-refresh. The model materializes successfully, the MV trigger is created, but **system.projections / system.data_skipping_indices** is empty for the target table.

Root cause is in `clickhouse__create_target_table` (/macros/materializations/materialized_view.sql:278):

```sql
{% macro clickhouse__create_target_table(relation, sql, catchup=True) -%}
  {% if catchup == True %}
    {{ get_create_table_as_sql(False, relation, sql) }}    {# → clickhouse__create_table_as → add_index_and_projections ✓ #}
  {% else %}
    {{ log('Catchup config set to false, skipping table backfill for ' + relation.name) }}
    {% set has_contract = config.get('contract').enforced %}
    {{ create_table_or_empty(False, relation, sql, has_contract) }}    {# → never calls add_index_and_projections ✗ #}
  {% endif %}
{%- endmacro %}
```

The **catchup=true** path dispatches to `clickhouse__create_table_as` (table.sql:266), which interleaves CREATE TABLE … EMPTY AS → add_index_and_projections(relation) → INSERT INTO. The catchup=false path calls create_table_or_empty directly, skipping the add_index_and_projections step entirely.

This affects all three call paths of clickhouse__create_target_table:
- First-time create (line 197)
- --full-refresh (line 217)
- clickhouse__replace_mv (line 441)

### Steps to reproduce

1. Create an MV model with both catchup=false and a projection:

```sql
{{ config(
    materialized='materialized_view',
    catchup=false,
    schema='mydb',
    order_by='(user_id, event_date)',
    engine='AggregatingMergeTree()',
    partition_by='toYYYYMM(event_date)',
    projections=[{
        'name': 'by_event_date',
        'query': 'SELECT user_id, event_date, uniq_visits_state ORDER BY (event_date, user_id)'
    }]
) }}
```

```sql
SELECT user_id,
       toDate(event_time) AS event_date,
       uniqStateIf(recipient_id, event_name = 'view') AS uniq_visits_state
FROM {{ source('eventdb', 'user_page_events') }}
GROUP BY user_id, event_date
```

2. Run dbt run --select <model> --target dev.
3. Query SELECT name FROM system.projections WHERE database='<schema>' AND table='<model>' — returns empty.
4. For comparison, change catchup=false → catchup=true, drop the target, re-run — projection appears.

### Expected behaviour

add_index_and_projections(relation) should be called regardless of the catchup setting. The data-load step is what catchup controls; projection/index DDL is a schema concern that's orthogonal to whether historical rows are backfilled.

Minimal in-tree workaround (project-level macro override at macros/clickhouse__create_target_table.sql) — mirrors the upstream catchup=true pattern but skips the INSERT:

```sql
{% macro clickhouse__create_target_table(relation, sql, catchup=True) -%}
  {% if catchup == True %}
    {{ get_create_table_as_sql(False, relation, sql) }}
  {% else %}
    {{ log('Catchup config set to false, skipping table backfill for ' + relation.name) }}
    {% set has_contract = config.get('contract').enforced %}
    {%- if adapter.is_before_version('22.7.1.2484') -%}
      {{ create_table_or_empty(False, relation, sql, has_contract) }}
    {%- else -%}
      {% call statement('create_table_empty') %}
        {{ create_table_or_empty(False, relation, sql, has_contract) }}
      {% endcall %}
      {{ add_index_and_projections(relation) }}
      select 1
    {%- endif -%}
  {% endif %}
{%- endmacro %}
```

The select 1 is a no-op placeholder for the outer {% call statement('main') %} wrapper (same pattern used upstream at materialized_view.sql:228). Suggested upstream fix: refactor clickhouse__create_target_table's catchup=false branch to use the same CREATE EMPTY → add_index_and_projections side-effect pattern as clickhouse__create_table_as.

### dbt and/or ClickHouse server logs

After `dbt run --select my_model` with `catchup=false`, querying **system.query_log** shows only two statements emitted:

1. CREATE TABLE mydb.my_model … EMPTY AS (SELECT …)
2. CREATE MATERIALIZED VIEW mydb.my_model_mv TO mydb.my_model AS …

No ALTER TABLE … ADD PROJECTION … is emitted. With catchup=true, three statements are emitted (the same two plus the INSERT INTO), and projections appear correctly because add_index_and_projections runs as a side effect inside clickhouse__create_table_as.

#### Environment

- dbt version: 1.11.7
- dbt-clickhouse version: 1.10.0
- clickhouse-connect version: 0.15.1
- Python version: 3.12.8
- Operating system: macOS (also reproduced on Linux CI)

#### ClickHouse server

- ClickHouse Server version: 26.2.10.10
- ClickHouse Server non-default settings, if any: none relevant
- CREATE TABLE statements for tables involved: any AggregatingMergeTree source/target — bug is materialization-layer, not schema-dependent
- Sample data: not required (the bug reproduces with an empty source table)

sankalpsthakur:
## Summary

Fixes #637

`clickhouse__create_target_table` with `catchup=false` called `create_table_or_empty` directly and never ran `add_index_and_projections`. The catchup flag is only meant to skip historical row backfill; projection and index DDL are schema concerns and must still apply on first create and full refresh.

### Changes
- For `catchup=false` on supported server versions: `CREATE TABLE … EMPTY` → `add_index_and_projections` → `select 1` (for the outer statement wrapper), matching the structure of `clickhouse__create_table_as` without the insert.
- Integration regression: MV with `catchup=false` and a projection config; assert the projection exists on the empty target table.

Independent of #689 (contract `query_settings` introspection); both are dbt-clickhouse bugfixes from the same batch.

## Checklist


sankalpsthakur:
related: #616

dbt/include/clickhouse/macros/materializations/materialized_view.sql
```diff
@@ -297,7 +297,18 @@
...
+       (projections, indexes) must still run. Fixes #637. #}
+    {% if adapter.is_before_version('22.7.1.2484') -%}
+      {{ create_table_or_empty(False, relation, sql, has_contract) }}
+    {%- else -%}
+      {% call statement('create_table_empty') %}
+        {{ create_table_or_empty(False, relation, sql, has_contract) }}
+      {% endcall %}
+      {{ add_index_and_projections(relation) }}
+      select 1
+    {%- endif -%}
```
koletzilla: Thanks @sankalpsthakur for the PR. A few suggestions:
- Please remove the part of the comment about `catchup only controls whether historical rows are inserted; schema DDL (projections, indexes) must still run. Fixes #637`. It doesn't add needed context.
- Remove the `adapter.is_before_version` part. That version is not maintained so we don't need to keep compatibility with it.
- Can you verify if the `select 1` is needed? feels like it can be skipped.

Apart from this, not needed for this PR: I don't like the idea of having this table logic repeated here and inside the `clickhouse__create_table_as` function as this bug may happen again in the future. But I have been checking and deduplicating that may carry more changes, so I'm ok doing that myself in a future PR.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
...
 * Fix multiple materialized views breaking when a `--` comment precedes the first marker. The view-name regex matched across newlines, so a leading comment was pulled into the marker, blanking the view query (ClickHouse `SYNTAX_ERROR`), or silently dropping the first MV when several are defined ([#682](https://github.com/ClickHouse/dbt-clickhouse/issues/682), [#683](https://github.com/ClickHouse/dbt-clickhouse/pull/683)).
+* Materialized view targets with `catchup=false` now create configured projections and indexes. Previously only the catchup=true path called `add_index_and_projections` ([#637](https://github.com/ClickHouse/dbt-clickhouse/issues/637)).
```
koletzilla: Please include the link to the PR instead of to the issue

koletzilla (changes requested):

koletzilla:
bugbot run

sankalpsthakur:
comment trimmed and is_before_version dropped in fec10ea. changelog points at the pr now.

kept the `select 1` — both call-statement blocks above it render to nothing, so it's the only body the enclosing `statement('main')` gets. same reason as the existing one at line 243. rendered it through macro_harness to check.

the red cloud job is a connection refused at setup on test_alias_model_name, unrelated to this pr.

sankalpsthakur:
rebased on main in 056e040 — #689 landing put a changelog conflict on this, both entries kept. no code change beyond that.

koletzilla:
Ey @sankalpsthakur, thanks again for making all the changes.

I still wasn't really sure about leaving the `SELECT 1` there so I have jumped it to find a solution and I have implemented it. The rest of the changes looks good to me!

I'll merge it once the tests pass

koletzilla (approved):
Thanks again for the PR!

PR #692, opened 2026-07-27
fix(snapshot): drop orphaned __dbt_tmp before recreating staging table

### LINKED ISSUE #691: Snapshot fails with TABLE_ALREADY_EXISTS on orphaned __dbt_tmp staging table
## Environment
- dbt-clickhouse 1.10.x
- Reproduced on a cluster with `ON CLUSTER`; the bug is topology-independent (affects single-node too)

## Symptom
`dbt snapshot` fails with:
```
Code: 57. DB::Exception: ... TABLE_ALREADY_EXISTS
```
on `create table <database>.<snapshot>__dbt_tmp`.

Manually running `DROP TABLE <snapshot>__dbt_tmp` makes the next run succeed once, but the failure recurs on a later run.

## Root cause
`build_snapshot_staging_table` in `dbt/include/clickhouse/macros/materializations/snapshot.sql` issues a plain `CREATE TABLE` for the `__dbt_tmp` staging relation - no pre-drop, no `IF [NOT] EXISTS`, no `OR REPLACE`:

```jinja
{% macro build_snapshot_staging_table(strategy, sql, target_relation) %}
    {% set tmp_relation = make_temp_relation(target_relation) %}
    {% set select = snapshot_staging_table(strategy, sql, target_relation) %}
    {% call statement('build_snapshot_staging_relation') %}
        {{ create_table_as(False, tmp_relation, select) }}
    {% endcall %}
    {% do return(tmp_relation) %}
{% endmacro %}
```

The **only** cleanup of `__dbt_tmp` is in `clickhouse__post_snapshot`, which runs at the very end of a *successful* run:

```jinja
{% macro clickhouse__post_snapshot(staging_relation) %}
    {{ drop_relation_if_exists(staging_relation) }}
{% endmacro %}
```

So any run that dies after `build_snapshot_staging_relation` but before `post_snapshot` (a statement error, an interrupt, a failing custom `post_hook`, etc.) **orphans `__dbt_tmp`**, and every subsequent run then crashes on the recreate - with no self-healing.

The `table` and `incremental` materializations don't have this gap because they call `drop_relation_if_exists(preexisting_intermediate_relation)` *before* creating the temp table. The snapshot materialization was written without that guard.

## Reproduce
1. Run a snapshot.
2. Cause it to fail after the staging table is created (e.g. kill the process mid-run, or make a downstream statement error).
3. Run `dbt snapshot` again → `TABLE_ALREADY_EXISTS` on `<snapshot>__dbt_tmp`. Repeatable indefinitely until the table is dropped by hand.

## Suggested fix
Add an idempotent pre-drop at the top of `build_snapshot_staging_table`:

```jinja
{% macro build_snapshot_staging_table(strategy, sql, target_relation) %}
    {% set tmp_relation = make_temp_relation(target_relation) %}

    {{ drop_relation_if_exists(tmp_relation) }}

    {% set select = snapshot_staging_table(strategy, sql, target_relation) %}
    {% call statement('build_snapshot_staging_relation') %}
        {{ create_table_as(False, tmp_relation, select) }}
    {% endcall %}
    {% do return(tmp_relation) %}
{% endmacro %}
```

`drop_relation_if_exists` → `clickhouse__drop_relation` (`dbt/include/clickhouse/macros/adapters.sql`) emits `DROP TABLE IF EXISTS <rel> <ON CLUSTER SYNC>`. `IF EXISTS` makes it a safe no-op when nothing's left, and the existing `SYNC` modifier already covers the on-cluster / replicated case. It uses the same `tmp_relation` object the create uses, so the `should_on_cluster` flag is identical between drop and create - no flag mismatch.

This makes snapshots self-healing: even if a run fails mid-flight and orphans `__dbt_tmp`, the next run drops it automatically instead of crashing.

## Related (same class of bug — not in scope of the proposed fix)
`clickhouse__snapshot_merge_sql` creates the `__snapshot_upsert` table with `CREATE TABLE IF NOT EXISTS` and only drops it at the very end. An orphaned upsert table is silently *reused and appended to* on retry.

bartcode:
## Summary

Closes #691.

`dbt snapshot` fails with `TABLE_ALREADY_EXISTS` on `<snapshot>__dbt_tmp` after any run that dies mid-snapshot. Reproduced on a cluster with `ON CLUSTER` (the bug is topology-independent).

**Root cause:** `build_snapshot_staging_table` in `dbt/include/clickhouse/macros/materializations/snapshot.sql` issued a plain `CREATE TABLE` for the `__dbt_tmp` staging relation — no pre-drop, no `IF [NOT] EXISTS`, no `OR REPLACE`. The only cleanup of `__dbt_tmp` is in `clickhouse__post_snapshot`, which runs at the end of a *successful* run. So any run that died after `build_snapshot_staging_relation` but before `post_snapshot` (a statement error, an interrupt, a failing custom `post_hook`, …) orphaned `__dbt_tmp`, and every subsequent run then crashed on the recreate — with no self-healing. The `table` / `incremental` materializations don't have this gap because they call `drop_relation_if_exists(preexisting_intermediate_relation)` *before* creating the temp table; the snapshot materialization was written without that guard.

**Fix:** one idempotent pre-drop at the top of `build_snapshot_staging_table`:

```jinja
{% set tmp_relation = make_temp_relation(target_relation) %}

{{ drop_relation_if_exists(tmp_relation) }}
```

`drop_relation_if_exists` → `clickhouse__drop_relation` emits `DROP TABLE IF EXISTS <rel> <ON CLUSTER SYNC>` (`adapters.sql`). `IF EXISTS` makes it a safe no-op when nothing's left, and the existing `SYNC` modifier already covers the on-cluster / replicated case. It uses the same `tmp_relation` object the create uses, so the `should_on_cluster` flag is identical between drop and create — no flag mismatch. This makes snapshots self-healing: an orphaned `__dbt_tmp` from a failed run is dropped automatically on the next run instead of crashing.

## Test

Adds `tests/integration/adapter/basic/test_snapshot_staging_tmp_cleanup.py`, extending dbt's `BaseSnapshotTimestamp` (reuses its seed/snapshot fixtures). The test runs a snapshot, manually creates an orphaned `ts_snapshot__dbt_tmp` (simulating a crashed prior run), re-runs the snapshot, and asserts it succeeds.

### Verified locally ✅

Ran against the bundled `tests/integration/docker-compose.yml` 3-node cluster (`test_shard`, `ON CLUSTER`), ClickHouse `latest`:

```
DBT_CH_TEST_USE_DOCKER=True
DBT_CH_TEST_CLUSTER=test_shard
pytest tests/integration/adapter/basic/test_snapshot_staging_tmp_cleanup.py
```

Result: **2 passed in 28.26s** — the inherited `test_snapshot_timestamp` (happy path) and the new `test_orphaned_staging_tmp_is_dropped_before_recreate` (regression).

**Confirmed the new test is a genuine regression test** — with the fix reverted, it fails with the exact error from the issue:

```
Code: 57. DB::Exception: Table ...ts_snapshot__dbt_tmp already exists. (TABLE_ALREADY_EXISTS)
```

`make lint` passes (black / isort / mypy / yamllint).

## Checklist
- [x] Integration test covering the scenario was added
- [x] A human-readable description of the changes was provided in CHANGELOG
- [ ] For significant changes, documentation in https://github.com/ClickHouse/clickhouse-docs was updated with further explanations or tutorials — N/A (bug fix, no user-facing surface change)

## Notes
- Out of scope (flagged in the issue, same class of bug): `clickhouse__snapshot_merge_sql` creates the `__snapshot_upsert` table with `CREATE TABLE IF NOT EXISTS` and only drops it at the end, so an orphaned upsert table is silently *reused and appended to* on retry — a data-correctness risk. Happy to address in a follow-up if maintainers want it.


koletzilla:
bugbot run

koletzilla (approved):

PR #697, opened 2026-08-02
fix: allow cluster profiles without database engine

fallintoplace:
## Summary
- normalize an omitted `database_engine` before evaluating cluster support
- add parameterized coverage for empty, standard, and replicated database engines

## Details
`database_engine` is optional in profile credentials. When a cluster was configured without it, the adapter passed `None` to the cluster helper, which then called `.lower()` and stopped compilation.

## Testing
- `pytest tests/unit -q`
- Ruff format, lint, and mypy checks for changed files
- YAML lint


koletzilla:
Hi @fallintoplace thanks for your PRs!

Can you please:
- Add a changelog to each of them similar as the rest of the rest of changelogs entries. As these are small fixes, please keep each changelog small.
- Rebase your branch on top of latest main

koletzilla (approved):

koletzilla:
Thanks for the PR!

Issue #666, opened 2026-06-08
Fix the new set of Flaky tests randomly failing in CI


koletzilla:
We are having several errors related to random tests failing. I have divided the failed tests in two sections:

### Tests that just need to wait until the data is replicated
 Seems like in all these situations the problem is that the test is asserting that some data needs to be present but it request it to a replica that have not them yet. The main fix for them would be to retry the check until the data is replicated:

FAILED tests/integration/adapter/incremental/test_base_incremental.py::TestIncrementalCompoundKey::test_compound_key - assert 200 == 180
- https://github.com/ClickHouse/dbt-clickhouse/actions/runs/27113511086/job/80015787918

FAILED tests/integration/adapter/materialized_view/test_materialized_view.py::TestCatchup::test_full_refresh_catchup_enabled - assert 3 == 4
- https://github.com/ClickHouse/dbt-clickhouse/actions/runs/26927605490/job/79440668311
- https://github.com/ClickHouse/dbt-clickhouse/actions/runs/26861082807/job/79214569859
- https://github.com/ClickHouse/dbt-clickhouse/actions/runs/26795644122/job/78991245933
- https://github.com/ClickHouse/dbt-clickhouse/actions/runs/26732836076/job/78780023484

FAILED tests/integration/adapter/incremental/test_schema_change.py::TestOnSchemaChange::test_append[schema_change_append] - assert 4 == 5
- https://github.com/ClickHouse/dbt-clickhouse/actions/runs/26615312343/job/78429598535

FAILED tests/integration/adapter/clickhouse/test_clickhouse_table_materializations.py::TestMergeTreeTableMaterialization - assert 10 >= 20
- https://github.com/ClickHouse/dbt-clickhouse/pull/639

### CH error codes that we need to retry
There are situations where one operation may fail because CH cannot process it in this particular moment. If the error is retryable, we should retry until the operation success.

Received ClickHouse exception, **code: 244**, server response: Code: 244. DB::Exception: Got unexpected ZooKeeper error ZNODEEXISTS (at index 10) for part all_0_0_0. (UNEXPECTED_ZOOKEEPER_ERROR) (for url https://***:8443)
- FAILED tests/integration/adapter/basic/test_ephemeral.py::TestEphemeral::test_ephemeral - AssertionError: dbt exit state did not match expected https://github.com/ClickHouse/dbt-clickhouse/actions/runs/27080994865/job/79926683096
[...]

PR #715, opened 2026-08-13
Reduce test flakyness: wait for data replication

koletzilla:
## Summary
Related tests mentioned in https://github.com/ClickHouse/dbt-clickhouse/issues/666

Fixes:
- Correctly set `select_sequential_consistency` for cloud tests: it's added when the config `dbt-engine: Shared` but that config wasn't there 🤦 Adding it now. 
  - Fixes test like https://github.com/ClickHouse/dbt-clickhouse/actions/runs/26615312343/job/78429598535 `FAILED tests/integration/adapter/incremental/test_schema_change.py::TestOnSchemaChange::test_append[schema_change_append] - assert 4 == 5` where another replica didn't have yet all the data, now it will had it 
- Fix https://github.com/ClickHouse/dbt-clickhouse/actions/runs/32270026376/job/96123916802?pr=715 `tests/integration/adapter/projections/test_projections.py::TestProjections::test_create_and_verify_multiple_projections - assert 0 > 0`: query_log is local in Cloud, so a query reading a local query_log may be just reading from the wrong replica. Added ON CLUSTER for both operations
- Set `lightweight_deletes_sync=3` in Cloud so it only waits for active replicas. Doesn't fix an specific issue now, but it's better to have it because something will break eventually.

- Fix https://github.com/ClickHouse/dbt-clickhouse/actions/runs/32273552665/job/96135475425?pr=715 `tests/integration/adapter/incremental/test_base_incremental.py::TestIncrementalCompoundKey::test_compound_key - assert 280 == 260`: subquery may select data that has not been replicated yet. Fixing it by adding on Shared database engines the setting `select_sequential_consistency` inside the subquery

### DIFF
--- a/dbt/adapters/clickhouse/dbclient.py
+++ b/dbt/adapters/clickhouse/dbclient.py
@@ -96,6 +96,7 @@ class ChClientWrapper(ABC):
         if credentials.database_engine == 'Shared':
             self._conn_settings.setdefault('select_sequential_consistency', '1')
         self._conn_settings.setdefault('mutations_sync', '3')
+        self._conn_settings.setdefault('lightweight_deletes_sync', '3')
         self._conn_settings.setdefault('alter_sync', '3')
         self._conn_settings.setdefault('insert_distributed_sync', '1')
         self._client = self._create_client(credentials)
--- a/tests/integration/adapter/projections/test_projections.py
+++ b/tests/integration/adapter/projections/test_projections.py
@@ -147,15 +147,19 @@ class TestProjections:
             % "table",
         }
 
+    def _get_cluster(self) -> str:
+        # In Cloud we don't need to use `ON CLUSTER` for regular operations, but it's still useful for some
+        # edge cases like flushing/querying all the different query_log.
+        cluster = os.environ.get("DBT_CH_TEST_CLUSTER", "").strip()
+        on_cloud = os.environ.get("DBT_CH_TEST_CLOUD", "").lower() in ("1", "true", "yes")
+        return "default" if not cluster and on_cloud else cluster
+
     def _get_table_reference(self, table: str) -> str:
-        return (
-            table
-            if os.environ.get("DBT_CH_TEST_CLUSTER", "").strip() == ""
-            else f"clusterAllReplicas({os.environ.get('DBT_CH_TEST_CLUSTER')}, {table})"
-        )
+        cluster = self._get_cluster()
+        return table if not cluster else f"clusterAllReplicas('{cluster}', {table})"
 
     def _flush_system_logs(self, project) -> None:
-        cluster = os.environ.get("DBT_CH_TEST_CLUSTER", "").strip()
+        cluster = self._get_cluster()
         cluster_clause = f'ON CLUSTER "{cluster}"' if cluster else ""
         project.run_sql(f"SYSTEM FLUSH LOGS {cluster_clause}", fetch="all")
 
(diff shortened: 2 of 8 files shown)


BentsiLeviav:
@koletzilla, what other stronger solutions can we utilize to achieve consistency? Any settings (like select_sequential_consistency)? Maybe a sticky replica for tests?

koletzilla:
> @koletzilla, what other stronger solutions can we utilize to achieve consistency? Any settings (like select_sequential_consistency)? Maybe a sticky replica for tests?

Really good question. I thought we were already using things like `select_sequential_consistency` as I already added that by default when the adapter was configured with `db-engine=Shared` but... the cloud tests weren't configured with that, so the config wasn't used 🤦 

So some tests may still need to retry until something happen as some operations may not be "sync": MV refresh, query_log flush... but for some of them, as the ones I was retrying now, Looks like correctly setting the expected `select_sequential_consistency` may be enough to avoid them.

I have also added it to the internals of the lightweight delete operations as these may be impacted as well.

Added also `lightweight_deletes_sync=3 ` because that would be needed for other flakys

>  Maybe a sticky replica for tests?
Connections are already done against a particular replica, so I don't expect us from needing this. Let's see how these fixes work and then iterate with that if we see we don't have more options.

dbt/adapters/clickhouse/dbclient.py
```diff
@@ -96,6 +96,7 @@ def __init__(self, credentials: ClickHouseCredentials):
         if credentials.database_engine == 'Shared':
             self._conn_settings.setdefault('select_sequential_consistency', '1')
         self._conn_settings.setdefault('mutations_sync', '3')
+        self._conn_settings.setdefault('lightweight_deletes_sync', '3')
```
mzitnik, line 99: lightweight_deletes_sync = 3 
What does it mean where can i find a refernce for that?

tests/integration/adapter/projections/test_projections.py
```diff
@@ -147,15 +147,19 @@ def models(self):
             % "table",
         }
 
+    def _get_cluster(self) -> str:
+        # In Cloud we don't need to use `ON CLUSTER` for regular operations, but it's still useful for some
+        # edge cases like flushing/querying all the different query_log.
+        cluster = os.environ.get("DBT_CH_TEST_CLUSTER", "").strip()
+        on_cloud = os.environ.get("DBT_CH_TEST_CLOUD", "").lower() in ("1", "true", "yes")
```
mzitnik, line 154: if this repet it self in more place lets consider having a helper function

dbt/adapters/clickhouse/dbclient.py
koletzilla: https://clickhouse.com/docs/reference/settings/session-settings/lightweight#lightweight_deletes_sync

The default value is already `2`, Changing it to 3 in the same way we did for `mutations_sync` and `alter_sync`, that way the queries don't wait for replicas that aren't active, like the ones that are shutting down.

We already applied the `3` value also in Fivetran for `mutations_sync` and `alter_sync`

tests/integration/adapter/projections/test_projections.py
koletzilla: I thought it was but other parts of the code follows this logic slightly different: In other parts it justs takes the value of `DBT_CH_TEST_CLUSTER` as cluster name.

In this case, we want to specify a cluster name always, even in situations where `DBT_CH_TEST_CLUSTER` is not set, as this is one of the few moments where `ON CLUSTER` needs to be used in a cloud env so we can call `SYSTEM FLUSH LOGS` on all nodes and query all query_logs at the same time. For the rest of the code, `ON CLUSTER` should only be used if `DBT_CH_TEST_CLUSTER` is set

mzitnik (approved):

